PR #26, opened 2015-02-09
Download files to temporary directory first.

### LINKED ISSUE #23: Do not overwrite destination while downloading
Currently the destination file becomes size 0 bytes while the download is in progress. Please consider using a temp file for downloading and only overwrite the destination when the download was successful and checksum matches.

This would suit our scenario where we download a war and overwrite a previous version in the webapps directory.

nanliu:
Fix #23

### DIFF
--- a/lib/puppet/provider/archive/default.rb
+++ b/lib/puppet/provider/archive/default.rb
@@ -9,6 +9,8 @@ rescue LoadError
   require File.join archive.path, 'lib/puppet_x/bodeco/util'
 end
 
+require 'tempfile'
+
 Puppet::Type.type(:archive).provide(:default) do
   attr_reader :archive_checksum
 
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/archive/default.rb
```diff
@@ -40,6 +41,20 @@ def archive_filepath
     resource[:path]
   end
 
+  def download(archive_filepath)
+    tempfile=Tempfile.new(resource[:name])
+    PuppetX::Bodeco::Util.download(resource[:source], tempfile.path, :username => resource[:username], :password => resource[:password], :cookie => resource[:cookie] )
+    tempfile.close
```
elyscape: Be advised that this may delete the file if `tempfile` is finalized before you move it elsewhere.  I would recommend replacing lines 45-47 with this:
```ruby
tempfile=Tempfile.new(resource[:name])
tempfile.close!
PuppetX::Bodeco::Util.download(resource[:source], tempfile.path, :username => resource[:username], :password => resource[:password], :cookie => resource[:cookie] )
```
`Tempfile#close!` forces deletion of the file to happen immediately.

nanliu: Thanks for the review, I've added your recommended fix in #28.

PR #27, opened 2015-02-10
Support configuring extract command user/group.

### LINKED ISSUE #22: Support owner / group file permissions
In nanliu-staging, it was possible to set the owner and group when deploying an archive.  I couldn't find similar support in nanliu-archive.  Could this support be added?

nanliu:
In order to extract the file, we need to manage the file permission.
Setting file ownership is requires uid/gid and not the username, and
some serious complexity in Windows. This will do for now.

Fix #22

### DIFF
--- a/lib/puppet/provider/archive/default.rb
+++ b/lib/puppet/provider/archive/default.rb
@@ -96,7 +96,14 @@ Puppet::Type.type(:archive).provide(:default) do
   def extract
     if resource[:extract] == :true
       raise(ArgumentError, "missing archive extract_path") unless resource[:extract_path]
-      PuppetX::Bodeco::Archive.new(archive_filepath).extract(resource[:extract_path], nil, resource[:extract_flags])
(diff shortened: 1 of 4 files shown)


lib/puppet/provider/archive/default.rb
```diff
@@ -96,7 +96,14 @@ def cleanup
   def extract
     if resource[:extract] == :true
       raise(ArgumentError, "missing archive extract_path") unless resource[:extract_path]
-      PuppetX::Bodeco::Archive.new(archive_filepath).extract(resource[:extract_path], nil, resource[:extract_flags])
+      PuppetX::Bodeco::Archive.new(archive_filepath).
+        extract(
+          resource[:extract_path],
+          :custom_command => resource[:extract_command],
```
myoung34: I think this causes a syntax error, you're passing hash key/values as params instead of a hash

PR #34, opened 2015-02-25
disable travis.ci sudo.

nanliu:


elyscape:
This doesn't actually add 1.8.7 to the matrix.

nanliu:
The original commit added 1.8.7 to the travis matrix (since they are in maintenance I can't point to the error), but I'm unable to make it run due to gem requirements for the testing suite. I'll do a best effort support for ruby 1.8.7.

elyscape:
Makes sense.

PR #41, opened 2015-04-10
Issue #45 - set the namevar to be full path to file

JoshiiSinfield:
This resolves the duplicate alias for the resources by not stripping the
directory off for the namevar. 
Stripping it off during the creation of the tempfile keeps current
functionality for the tempfile creation.


JoshiiSinfield:
I'll have a crack at fixing the tests, and I'll try put some new ones in too.

nanliu:
See #47.

Issue #43, opened 2015-04-22
'7zip' variables incompatible with Puppet 4.0


antoineco:
With the new parser introduced in Puppet 4.0 variable names are no allowed to start with a digit.

nanliu:
Ok, looks a like I have a few bugs that are breaking changes. I'll create another branch and queue them up for a release.

nanliu:
Thanks for the PR, this will be resolved in #45.

nanliu:
Please review and test 0.3.x branch.

PR #44, opened 2015-04-22
fix 7zip var names

antoineco:
Puppet 4.0 compatibility (#43)


nanliu:
This will be resolve as part of #45. Thanks for the PR.

Issue #50, opened 2015-04-25
Sonatype Nexus support (archive::nexus)


jairojunior:
* You can Query Nexus HTTP API to download an artifact, e.g.:

https://oss.sonatype.org/service/local/artifact/maven/content?g=io.hawt&a=hawtio-web&v=1.4.36&p=war&r=releases

* And with a little trick you can also download md5 or sha1:

https://oss.sonatype.org/service/local/artifact/maven/content?g=io.hawt&a=hawtio-web&v=1.4.36&p=war.md5&r=releases (p=war.md5 or war.sha1)

I was playing around trying to create a simple implementation for this using archive: https://github.com/jairojunior/puppet-archive/blob/master/manifests/nexus.pp

But Bodeco.Util.download doesn't support query strings. Any chances to add support for query string in this module? Or it's not intended to this (could break something or any other reason that I can't see)?

I'm really interested to add this support to archive, so any help would be appreciated.

nanliu:
The other two define types use a function because the checksum file needs processing. The archive parameter checksum_source was intended for this, but not yet implemented. I think checksum_url might be a better name?

jairojunior:
@nanliu If you're working with a remote source (http, ftp, s3), checksum_url is probably a better name.

nanliu:
@jairojunior please review latest changes and provide feedback if you run into any issues.

nanliu:
Sorry didn't mean to close issue. Per last comment, I will accept a nexus define type wrapper. I'll list it as experimental along with the go/artifactory wrapper in the readme.

jairojunior:
Thank you. 

I was able to finish archive::nexus with the latest changes, but I have a small concern about performance.

Check this gist: https://gist.github.com/jairojunior/071a5256abeb1d39e54a

In the second run, with HTTP checksum URL it tooks 12s, with HTTPS, 6s, and hardcoded checksum only takes 0.4s

I have a module that uses net/http to query a HTTP API and the request are really fast (~100ms). Do you have any idea of what is causing this?

Meanwhile, I'll work on rspec tests for archive::nexus.

nanliu:
I'm not able to reproduce the difference. Can you time curl to the url in comparison:
https://gist.github.com/nanliu/0d3f5bf4a53431f81ed2

jairojunior:
My bad, it's something with my puppetlabs/centos-6.6-64-puppet vagrant setup:

[...]

PR #52, opened 2015-04-27
Add support for remote checksum url

nanliu:
* Add support for url query for remote file.
* Fix type documentation.
* checksum_source was never used, so backwards compatible name change.


nanliu:
FIX #50

PR #54, opened 2015-04-28
Add archive::nexus support

jairojunior:
Still need to build tests, but I'm not sure whether you prefer a small or a big PR...


manifests/nexus.pp
```diff
@@ -0,0 +1,69 @@
+#
+# Download from Nexus using REST API
+# More info here: https://repository.sonatype.org/nexus-restlet1x-plugin/default/docs/path__artifact_maven_content.html
+#
+define archive::nexus (
+  $ensure       = present,
+  $checksum_type = 'md5',
+  $nexus_url    = undef,
+  $gav          = undef,
```
nanliu: Any specific reason why gav is crammed into one parameter?

jairojunior: It's gradle syntax for dependecy declaration:

http://mvnrepository.com/artifact/io.hawt/hawtio-web/1.4.49#gradle

Most developers says that it is more readable cause you find get all (important) package information in a single line.

The same is used here: https://github.com/cescoffier/puppet-nexus

And I'm not sure right now, but Nexus uses same GAV syntax when you're manually uploading an artifact.

What do you think? Declare each field (g, a, v)? Single field (gav)? Or support both?

nanliu: Oh, that's fine if it's nexus convention. You don't have to repeat nexus in the parameter, so it might be better just to leave it as archive::nexus::url instead of archive::nexus::nexus_url.

jairojunior:
@nanliu I refactored archive::nexus:nexus_url and did the tests, but it's failing when future_parser=yes.

It's probably because of split and regsubst functions, but I don't know how to solve this. I'm somewhat new in puppet/ruby ecosystem. Can you give me a hand again?

nanliu:
I suspect future_parser doesn't automatically map undef to empty string ''. One of your test did not supply the attribute [archive::nexus::packaging](https://github.com/puppet-community/puppet-archive/pull/54/files#diff-95cf2a228a8db88a536b98067b4ea6c4R38) and this will result in undef in the following line:

$checksum_url = regsubst($artifact_url, "p=${packaging}", "p=${packaging}.${checksum_type}")

The best way to debug is adding pry and see what it's barfing by stepping through the test. (Add pry gem to Gemfile and run bundle install).

```ruby
require 'pry'
binding.pry
```

spec/defines/nexus_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    let :params do
+      {
+        owner: 'tom',
+        group: 'worker',
+        user: 'tom',
+        extract: true,
+        extract_path: '/opt',
+        creates: '/opt/artifact/WEB-INF',
+        cleanup: true
+      }
```
nanliu: This is the test that's missing packaging attribute. I think some of the attributes defaults should be '' instead of undef.

jairojunior: gav and repository are required parameters, and packaging, when supressed, defaults to 'jar' (in Nexus API), so I made packaging defaults to 'jar' in order to produce the same effect.

Still getting the error. I'll investigate with pry.

elyscape: > gav and repository are required parameters

In that case, they shouldn't have defaults at all, even of `undef`.

manifests/nexus.pp
```diff
@@ -0,0 +1,68 @@
...
+  $group        = undef,
+  $extract      = undef,
+  $extract_path = undef,
+  $creates      = undef,
+  $cleanup      = undef,
+) {
+
+  include archive::params
+
+  $artifact_info = split($gav, ':')
```
elyscape: This line is what's causing problems with the future parser.  `split()` doesn't know how to handle `undef`.

jairojunior:
@nanliu @elyscape Thank you. I finally did. Sorry for the inconvenience.

PR #58, opened 2015-05-18
fix params.pp so that it works with strict_mode

robbyt:
Without this change, strict_mode will make puppet say:

`Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Undefined variable "archive::params::seven_zip_name"; Undefined variable "seven_zip_name" at /etc/puppet/environments/production/modules/archive/manifests/init.pp:14 on node puppet`


nanliu:
+1

nibalizer:
+1

PR #61, opened 2015-05-27
Added wget provider when using unix. 

### LINKED ISSUE #60: Remove the dependency on Faraday
For our use case it is difficult to get an extra gem installed. At the moment, this means we cannot use the module. I would like to use at least the types (and providers) in this module, but they are 'hard wired' using faraday.

I've looked at the code and it seems you are using faraday mostly as an easy way to call NET:HTTP. I could try and make a PR removing Faraday from the provider. Is this beneficial? Am I missing something in the reasoning for Faraday?

hajee:
Also extracted rspec's for archive providers. Righ now there is no specific test for either the wget provider, so it uses the normal one.

Renamed the default provider to ruby, because now we have more then one provider and this name is a better fit for what it's doing.

This PR closes #60 at least for unix systems containing `wget`.


nanliu:
Overall looks fine, we still need to migrate require 'puppet_x/bodeco/util' so it doesn't load faraday gem. I'm not sure we need shared context since as a parent provider methods don't need to be tested multiple times. If we implement spec it would be strictly for the download method for the wget provider.

lib/puppet/provider/archive/wget.rb
```diff
@@ -0,0 +1,42 @@
+Puppet::Type.type(:archive).provide(:wget, :parent => :ruby ) do
+
+  commands   :wget => '/usr/bin/wget'
+  defaultfor :operatingsystem => :linux
```
nanliu: I don't believe operatingsystem fact returns linux. This repo is pretty out of date, but would give a better idea what facts might be suitable:
https://github.com/apenney/puppet_facts/blob/master/PE2.8/redhat-5-i386.facts#L52

I believe the closest might be feature_match, which would be :feature => :posix:
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider.rb#L285

lib/puppet/provider/archive/wget.rb
```diff
@@ -0,0 +1,42 @@
+Puppet::Type.type(:archive).provide(:wget, :parent => :ruby ) do
+
+  commands   :wget => '/usr/bin/wget'
```
nanliu: Unless we must use /usr/bin/wget, it's better just to use wget and let puppet find the command in the path.

nanliu:
If we switch default provider for linux, not sure if curl or wget be the default provider, or should we implement it as a fact (similar to below)?

https://github.com/puppet-community/puppet-staging/blob/master/lib/facter/staging_http_get.rb

lib/puppet/provider/archive/wget.rb
hajee: You're right. I will change this.

hajee:
My reasoning was that if wget is not available, we fall back to the default implementation in ruby.

lib/puppet/provider/archive/wget.rb
hajee: The operatingsystem fact indeed doesn't return :linux. The strange thing is, I testen in on a linux system and it works. I could also check if it is **not** windows. What do you think.

hajee:
The code right now still needs the faraday gem when using the ruby provider. My reasons for using the shared behaviour was based on the idea that you could have more providers with all slightly different behaviour. 

If you think it is YAGNU, I'll remove it.

nanliu:
So content and download are the only method dependent on faraday, I think we can make net::http the default provider if we remove redirect as a requirement and make faraday another provider. This should allow migration:

```puppet
Archive {
  provider => 'faraday' 
}
```

I've listed the goals for 0.4.0 in #62.

lib/puppet/provider/archive/wget.rb
nanliu: I wonder if it just used the default ruby provider when you were testing. I don't see a way to test for negate in defaultfor, only for confine in the puppet provider code.

hajee: I used puppet --debug and saw the command. Also used the debugger to actually step though the code. Anyway.. I'll change it to:
```ruby
defaultfor :feature => :posix
```

or

hajee:
I've finished the request changes and some more. I noticed that wget doesn't handle `file:` urls. That was my use case so therefore I also added a curl provider. 

In #62 you listed the `faraday` provider. So I extracted the `faraday` specific stuff from the `ruby` provider into a `faraday` provider. To get a full functional ruby provider without faraday, you would have to build a robust `download` method. I'll leave that to someone else ;-)

Although both providers have a unit spec, the whole type and provider, could use a Beaker acceptance test. I guess this is an other thing for the roadmap.

nanliu:
We still need to migrate the content method and implement a naive net/http download for ruby. I'll merge this as is.

PR #66, opened 2015-07-02
allow setting of mode for artifacts pulled from nexus

smbambling:
Grant the ability to set the mode on artifacts retrieved from a nexus repository.


nibalizer:
lgtm

PR #69, opened 2015-08-11
Added redirect to  curl provide. Closes #65

### LINKED ISSUE #65: Follow redirect using curl
I needed curl to follow redirects, with -L parameter, i.e.:

```
curl http://google.com.br/ -o google.html
```
gets the wrong file, while:

```
curl http://google.com.br/ -L -o google.html
```
gets the actual index file.

Do you agree that the curl provider should work this way, following redirects? I made the changes on my fork but I want to know what do you think about it before making a pull request.

hajee:


smbambling:
:+1:

PR #76, opened 2015-10-19
Add support for bzip

craigmunro:
Solaris part is untested.


nanliu:
Could you add the corresponding tests? see https://github.com/puppet-community/puppet-archive/blob/master/spec/unit/puppet_x/bodeco/archive_spec.rb#L31-L33

PR #82, opened 2015-11-11
Modulesync

jyaworski:
Don't merge, yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,34 @@
 ---
+notifications:
+  email: false
+sudo: false
 language: ruby
-bundler_args: --without development
-script: "bundle exec rake validate && bundle exec rake lint && bundle exec rake spec SPEC_OPTS='--format documentation'"
+cache: bundler
+bundler_args: --without system_tests
+addons:
+  apt:
+    sources:
+      - augeas
+    packages:
+      - libaugeas-dev
+before_install: rm Gemfile.lock || true
+script:
+  - 'bundle exec rake $CHECK'
 matrix:
   fast_finish: true
   include:
   - rvm: 1.9.3
-    env: PUPPET_GEM_VERSION="~> 3.0"
+    env: PUPPET_VERSION="~> 3.4" CHECK=test
   - rvm: 1.9.3
-    env: PUPPET_GEM_VERSION="~> 3.0" FUTURE_PARSER="yes"
-  - rvm: 2.1.5
-    env: PUPPET_GEM_VERSION="~> 3.0"
-  - rvm: 2.1.5
-    env: PUPPET_GEM_VERSION="~> 3.0" FUTURE_PARSER="yes"
-sudo: false
-notifications:
-  email: false
-  flowdock: 330d85ed4bb7657fc334633f4b9209ee
+    env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.1.6
+    env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.1.6
+    env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.2.3
+    env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.2.3
+    env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=rubocop
 deploy:
   provider: puppetforge
   user: puppet
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,59 @@
-source "https://rubygems.org"
+source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
-group :development, :test do
-  gem 'bodeco_module_helper', :git => 'https://github.com/bodeco/bodeco_module_helper.git'
-  gem 'yard'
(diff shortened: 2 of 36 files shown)


jyaworski:
@igalic we should consider adding rspec-puppet-utils to the default gemfile. This module needs it.

jyaworski:
OK, we're good to go here. @igalic

igalic:
@jyaworski what does rspec-puppet-utils do, and why isn't it default yet?

nanliu:
@igalic rspec-puppet-utils have some helpers for hiera data testing. I'm not 100% sure what's in use in rspec testing in this module, but this seems like a much needed update.

lib/puppet/type/archive.rb
```diff
@@ -19,60 +19,60 @@
     defaultto(:present)
 
     # The following changes allows us to notify if the resource is being replaced
-    def is_to_s(value)
-      return "(#{resource[:checksum_type]})#{self.provider.archive_checksum}" if self.provider.archive_checksum
+    def s?(value)
```
nanliu: Should this stay 'def is_to_s', seems like an unintended change.

Gemfile
```diff
@@ -1,10 +1,34 @@
...
-  gem 'bodeco_module_helper', :git => 'https://github.com/bodeco/bodeco_module_helper.git'
-  gem 'yard'
-  gem 'faraday'
-  gem 'faraday_middleware'
+group :test do
+  gem 'rake'
+  gem 'puppet', ENV['PUPPET_VERSION'] || '~> 3.8.0'
+  gem 'rspec-puppet', git: 'https://github.com/rodjek/rspec-puppet.git'
+  gem 'rspec-puppet-augeas'
+  gem 'ruby-augeas'
```
nanliu: Not sure we need either augeas gem.

.travis.yml
```diff
@@ -1,22 +1,34 @@
...
-bundler_args: --without development
-script: "bundle exec rake validate && bundle exec rake lint && bundle exec rake spec SPEC_OPTS='--format documentation'"
+cache: bundler
+bundler_args: --without system_tests
+addons:
+  apt:
+    sources:
+      - augeas
+    packages:
+      - libaugeas-dev
```
nanliu: This can be removed since the module does not have any augeas dependencies.

lib/puppet/type/archive.rb
jyaworski: It was intentional. Rubocop complained.

Gemfile
jyaworski: The augeas gem is part of the default gemset that modulesync gives people. @igalic should we have a boolean to toggle this?

.travis.yml
jyaworski: See my other comment.

Gemfile
igalic: or remove it from the default gemset?

lib/puppet/type/archive.rb
nanliu: pretty sure we should ignore rubocop since this is overriding is_to_s method specified in ensurable which comes from property.rb:
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/property/ensure.rb#L57
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/property.rb#L349-L351

Gemfile
nanliu: +1 removal from default gemset.

jyaworski: I'll see if I can write up a patch, or if we can just use https://github.com/puppet-community/modulesync_config/pull/37 and put the onus on the people that use augeas to include the gems...

lib/puppet/type/archive.rb
jyaworski: I'll revert it. We'll see about ignoring this example in the base config.

jyaworski:
Hmm... this is somehow failing, when the file that travis fails on doesn't fail for me:

````
joseph.yaworski@jyaworski-mbpr ..dulesync_config/modules/puppet-archive (git)-[modulesync] % rvm 2.2.3 do rubocop spec/unit/puppet_x/bodeco/archive_spec.rb 
Inspecting 1 file
.

1 file inspected, no offenses detected
````

Could someone else look at this? Travis problem?

**Comment by nanliu** [comment]
It appears there's trailing whitespace on spec/unit/puppet_x/bodeco/archive_spec.rb line 34 in that commit. Maybe what you have locally is different?

**Comment by jyaworski** [comment]
@nanliu the interesting thing is that git status shows I'm current, and even if I try pushing it says I'm up-to-date.

**Comment by jyaworski** [comment]
No idea why it's failing now. Hmm....

**Comment by jyaworski** [comment]
This is finished finally. Let me figure out how to rebase all of this....

**Comment by jyaworski** [comment]
Hooray for starting completely over. This is ready to merge.

File: lib/puppet/type/archive.rb [hunk 4]
```diff
@@ -19,60 +19,60 @@
     defaultto(:present)
 
     # The following changes allows us to notify if the resource is being replaced
-    def is_to_s(value)
-      return "(#{resource[:checksum_type]})#{self.provider.archive_checksum}" if self.provider.archive_checksum
+    def is_to_s?(value) # rubocop:disable Style/PredicateName
```
nanliu: should not have ? at the end of the method, it should be def is_to_s(value)

jyaworski: Changed.

igalic:
@jyaworski ready for another sync?

jyaworski:
@igalic maybe. This one was kind of tough.

Issue #80, opened 2015-11-04
No way to set environment vars or a proxy for http


edestecd:
In staging module you can pass in an environment.  I use this to set http_proxy normally.
There is no equivalent.  Maybe this could be an option, if you continue to use Faraday.
https://github.com/lostisland/faraday/issues/221

PR #83, opened 2015-11-12
add proxy support, add new params to readme documentation

terrimonster:
Added proxy support to types and providers, as well as unit tests. I spun up a squid vagrant instance and a base instance, ran the tests/archive.pp with the new params set, and all ran well. Added the params to the readme documentation as well.

### DIFF
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -150,6 +150,15 @@ Puppet::Type.newtype(:archive) do
     desc 'extract command group (using this option will configure the archive file permisison to 0644 so the user can read the file).'
   end
 
+  newparam(:proxy_type) do
+    desc 'proxy type (none|ftp|http|https)'
+    newvalues(:none, :ftp, :http, :https)
+  end
+
+  newparam(:proxy_server) do
+    desc 'proxy address to use when accessing source'
+  end
+
   autorequire(:package) do
     'faraday_middleware'
   end
@@ -161,5 +170,10 @@ Puppet::Type.newtype(:archive) do
   validate do
     filepath = Pathname.new(self[:path])
     self[:filename] = filepath.basename.to_s
+    if self[:proxy_server]
+      self[:proxy_type] ||= URI(self[:proxy_server]).scheme.to_sym
+    else
+      self[:proxy_type] = :none
+    end
   end
 end
--- a/spec/unit/puppet/provider/archive/wget_spec.rb
+++ b/spec/unit/puppet/provider/archive/wget_spec.rb
@@ -78,5 +78,20 @@ RSpec.describe wget_provider do
         provider.download(name)
       end
     end
+
+    context 'proxy specified' do
+      let(:resource_properties) do
+        {
+          :name => name,
+          :source => 'http://home.lan/example.zip',
+          :proxy_server => 'https://home.lan:8080',
+        }
+      end
+
+      it 'calls wget with default options and header containing cookie' do
+        expect(provider).to receive(:wget).with(default_options << '--https_proxy=https://home.lan:8080')
+        provider.download(name)
+      end
+    end
   end
 end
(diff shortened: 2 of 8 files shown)


nanliu:
Could we just infer the proxy type from the proxy url?

terrimonster:
We probably could. I went the lazy route of leaving that up to the end user. I could code it in, however, if that is preferred.

nanliu:
Yeah, let's just infer it if it's not a big change. Also for the parameter, should we just simplify to 'proxy', any suggestions?

terrimonster:
That all can be done pretty easily. Give me a few.

jyaworski:
I think this was merged around the same time that #82 was. Can you rebase?

terrimonster:
Rebased, style-cleaned. I have kept in the proxy type and server. After thinking abotu this and discussions, it is risky to make assumptions and take the control out of the user's hands.

jyaworski:
Looks good. Can we rebase this down to one commit?

lib/puppet/type/archive.rb
```diff
@@ -150,6 +150,16 @@ def should_to_s(value)
     desc 'extract command group (using this option will configure the archive file permisison to 0644 so the user can read the file).'
   end
 
+  newparam(:proxy_type) do
+    desc 'proxy type (none|ftp|http|https)'
+    newvalues(:none, :ftp, :http, :https)
+    defaultto(:none)
```
nanliu: We can defaultto a sensible value based on proxy_server and let user override. Let me find you an example.

lib/puppet/type/archive.rb
```diff
@@ -150,6 +150,16 @@ def should_to_s(value)
...
+  end
+
+  newparam(:proxy_server) do
+    desc 'proxy address to use when accessing source'
+  end
+
```
nanliu: I think if you leave :proxy_type without any default you can at the end here do:
```ruby
validate
  self[:proxy_type] ||= URI(self[:proxy_server]).scheme if self[:proxy_server]
end
```

This should make it more convenient than forcing user to specify proxy_type every time.

terrimonster:
Added your suggestion, Nan. Thank you very much for the code block :)

jyaworski:
This looks great.

lib/puppet/type/archive.rb
nanliu: FYI with the new code please remove this defaultto

nanliu:
Ok, please remove the defaultto and squash commit, then we are ready to merge.

spec/unit/puppet/provider/archive/wget_spec.rb
```diff
@@ -78,5 +78,21 @@
         provider.download(name)
       end
     end
+
+    context 'proxy specified' do
+      let(:resource_properties) do
+        {
+          :name => name,
+          :source => 'http://home.lan/example.zip',
+          :proxy_type => 'https',
```
nanliu: Also with the new code we can rewrite this test without providing :proxy_type and see if it just do the right thing.

nanliu:
Fix #80.

nanliu:
Bla meant to merge and close.

lib/puppet/type/archive.rb
alexjfisher: @nanliu But is this correct?

`proxy_type` is used to set an environment variable such as `http_proxy` or `https_proxy`.  It should default to the scheme of `source` not the scheme of `proxy_server`?

nanliu: @alexjfisher, I believe so.

We are assuming if `proxy_server=https://foo_proxy` we are using `https` proxy_type, and we don't care if you are trying to download a file from `http://somefile`. Also the user can be explicit about the proxy_type and we won't guess.

alexjfisher: Still not sure I'm convinced.  But yeah, you can always override proxy_type. You may have guessed, but I was trying to debug https://github.com/voxpupuli/puppet-archive/issues/277

with wget on the command line for example
`http_proxy=http://proxy.example.com:3128 wget https://www.example.com/somefile`
the proxy in the environment variable `http_proxy` would not be used.

You'd have to do
`https_proxy=http://proxy.example.com:3128 wget https://www.example.com/somefile`
instead and the proxy/squid would tunnel the https request.  (Having it intercept https can be done but is *far* less common I think)

PR #85, opened 2015-11-17
Accept 128 char checksum for sha512.

nanliu:


nanliu:
Fix #83 since we didn't catch this when we fixed #8.

PR #86, opened 2015-11-18
Added support for puppet_gem required by the AIO agent

dylanratcliffe:
The AIO agent uses puppet_gem as its gem provider so we need to be able to deal with that

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -19,7 +19,7 @@ class archive::params {
 
   if $::puppetversion =~ /Puppet Enterprise/ and $::osfamily != 'Windows' {
     $gem_provider = 'pe_gem'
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -19,6 +19,8 @@
 
   if $::puppetversion =~ /Puppet Enterprise/ and $::osfamily != 'Windows' {
     $gem_provider = 'pe_gem'
+  } elsif pick($::aio_agent_version,false) {
```
nanliu: I don't believe the pick function is necessary, just elsif $::aio_agent_version

dylanratcliffe: Good point, I've been doing too much ruby lately :/

PR #89, opened 2015-11-21
Make the directory before moving the file there

### LINKED ISSUE #74: ensure directory exists before moving archive
When moving the archive after download we should create the directory structure so that the mv function does not fail.

https://github.com/puppet-community/puppet-archive/blob/0.3.0/lib/puppet/provider/archive/default.rb#L66

This can be fixed by inserting the following before the line above `    FileUtils.mkdir_p(File.dirname(archive_filepath))
`

I would issue a PR but since the next version is under a major refactor I was not sure if we wanted to create a 0.3.1 version.

jyaworski:
Fixes #74.

### DIFF
--- a/lib/puppet/provider/archive/curl.rb
+++ b/lib/puppet/provider/archive/curl.rb
@@ -45,6 +45,7 @@ Puppet::Type.type(:archive).provide(:curl, :parent => :ruby) do
       fail(Puppet::Error, 'Download file checksum mismatch') unless archive.checksum(resource[:checksum_type]) == checksum
     end
 
+    FileUtils.mkdir_p(File.dirname(archive_filepath))
     FileUtils.mv(temppath, archive_filepath)
   end
 end
--- a/lib/puppet/provider/archive/curl.rb
+++ b/lib/puppet/provider/archive/curl.rb
@@ -2,11 +2,7 @@ Puppet::Type.type(:archive).provide(:curl, :parent => :ruby) do
   commands :curl => 'curl'
   defaultfor :feature => :posix
 
-  def download(archive_filepath)
-    tempfile = Tempfile.new(tempfile_name)
-    temppath = tempfile.path
-    tempfile.close!
-
+  download do |temppath|
     @curl_params = [
       resource[:source],
       '-o',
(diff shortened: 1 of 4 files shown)


nanliu:
Since we are changing the same line in three different providers, seems like we should extract the common portions out so we aren't duplicating code in three different providers. Seems like in this case we should have download yield to each provider, so:

```ruby
  def download(archive_filepath)
    tempfile = Tempfile.new(tempfile_name)
    temppath = tempfile.path
    tempfile.close!

    yield temppath

    # conditionally verify checksum:
    if resource[:checksum_verify] == :true && resource[:checksum_type] != :none

      archive = PuppetX::Bodeco::Archive.new(temppath)
      fail(Puppet::Error, 'Download file checksum mismatch') unless archive.checksum(resource[:checksum_type]) == checksum
    end

    FileUtils.mkdir_p(File.dirname(archive_filepath))
    FileUtils.mv(temppath, archive_filepath)
  end
```

This way each provider can just implement the download portion:
```ruby
  download do |temppath|
    @curl_params = [
      resource[:source],
      '-o',
      temppath,
      '-L',
      '--max-redirs',
      5
    ]

    #
    # Manage username and password parameters
    #
    if resource[:username] && resource[:password]
      @curl_params << '--user' << "#{resource[:username]}:#{resource[:password]}"
    elsif resource[:username]
      @curl_params << '--user' << "#{resource[:username]}"
    elsif resource[:password]
      fail 'password specfied without username.'
    end

    if resource[:proxy_server]
      @curl_params << '--proxy' << "#{resource[:proxy_server]}"
    end

    #
    # Manage cookie parameter
    #
    @curl_params << '--cookie' << "#{resource[:cookie]}" if resource[:cookie]

    curl(@curl_params)
  end
```

jyaworski:
I'll look into it. I've never dealt with yield, but it makes sense after my initial pass. I just have to figure out where to put the download portion.

jyaworski:
All right; I actually have no idea how to implement that.

igalic:
@jyaworski need a hand? i can provide guidance in irc

nanliu:
looks like yield is unnecessary, and this would be easier to understand: https://github.com/nanliu/puppet-archive/commit/abd0ac3cde67d186e12bbeafc994c505205c55ec. I'll drop another PR.

nibalizer:
Fixed in #95

PR #90, opened 2015-11-25
Add authentication parameters to archive::nexus

kyblik:
Currently it is not possible to use archive::nexus with repos that require authentication. Also I added $extract_command, $proxy_server and $proxy_type to archive::nexus as they were also missing.


nanliu:
Please squash into a single commit.

kyblik:
Should be done.

PR #96, opened 2015-12-02
Update the ignore files to ensure builds work properly.

rnelson0:


igalic:
this whole thing needs moving into modulesync_config too plz

PR #104, opened 2015-12-06
Do not assume aio-agent if Puppet version >= 4

### LINKED ISSUE #103: puppet_gem provider broken on Fedora
Fedora packages do not install Puppet 4 in /opt but in standard OS paths, and rely on the OS Ruby instead of letting Puppet ship its own Ruby.

As a result the `puppet_gem` provider set in [params.pp](https://github.com/puppet-community/puppet-archive/blob/master/manifests/params.pp#L23) will not work because the [gem executable path set upstream](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/puppet_gem.rb#L15) does not exist.

I open this issue as a proposal to send a PR which fixes this. If supporting Fedora is not desired feel free to close the issue.

antoineco:
Fixes #103 

As far as I know the All-in-one agent is only available for Puppet 4, and having Puppet 4 installed doesn't necessarily mean it came with the All-in-one agent (OS like Fedora provide up-to-date rpm packages for Puppet 4)

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -19,9 +19,7 @@ class archive::params {
     }
   }
 
-  if versioncmp($::puppetversion, '4.0.0') >= 0 {
-    $gem_provider = 'puppet_gem'
-  } elsif $::puppetversion =~ /Puppet Enterprise/ and $::osfamily != 'Windows' {
+  if $::puppetversion =~ /Puppet Enterprise/ and $::osfamily != 'Windows' {
     $gem_provider = 'pe_gem'
   } elsif $::aio_agent_version {
     $gem_provider = 'puppet_gem'


rnelson0:
There is a fact is_pe that should probably be used rather than string
comparisons.
>
>

-- 

Rob Nelson
rnelson0@gmail.com

antoineco:
is_pe is not reliable on AIO agents: https://tickets.puppetlabs.com/browse/MODULES-2279

Please also note that the `aio_agent_version` fact was not set in my environment, so relying only on it might not be the best option. I'm currently trying to figure out if this is a bug or not: https://tickets.puppetlabs.com/browse/FACT-1280

nibalizer:
I think this is an ok solution. I'd like to see some facts show up in stdlib so that this code doesn't need to keep getting repeated.

antoineco:
Not sure if this would actually be accurate, looks like the fact I mentioned is only available on PE. It means there is currently no way to make the difference between AIO or not AIO on Puppet FOSS without implementing a custom fact.

eperdeme:
Stops my catalogues compiling due to missing fact.

Server - puppetserver 2.2.1-1-RHEL7
Client:  puppet     3.8.2-1puppetlabs1 

Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Unknown variable: '::aio_agent_version'. at /etc/puppetlabs/code/environments/develop/modules/archive/manifests/params.pp:26:11 at /etc/puppetlabs/code/environments/develop/modules/ui_broker/manifests/init.pp:33 on node test.node.com.
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run

antoineco:
@eperdeme I don't think it's related to this issue, the fact you mentioned is already present on master.

eperdeme:
The fact is not on my clients nor my puppet server.

[root@cxpxpos01mas001 /]# rpm -qa | grep puppetserver
puppetserver-2.2.1-1.el7.noarch
[root@cxpxpos01mas001 /]# facter -p 
2015-12-23 12:20:14.611425 WARN  puppetlabs.facter - skipping external facts for "/opt/puppetlabs/puppet/cache/facts.d": No such file or directory
[root@cxpxpos01mas001 /]#

The run works fine if on the agent I do "export FACTER_aio_agent_version=3.8.2".

In puppet4 the variable/fact has to exist or it breaks the puppet run and as the client does not have the fact for "aio_agent_version" in Puppet 3.8.x (or seemly in Opensource Puppetserver 2.x) it fails the run.

antoineco:
I just meant that you should open a new Issue to report your problem, this is a pull request which tries to solve something different from what you are describing.

igalic:
thanks @antoineco!

PR #111, opened 2016-01-19
Only use the faraday provider when the faraday_middleware gem is installed

epienbroek:
On systems where wget is installed, but curl isn't the following warning occurs:
```
  Warning: Found multiple default providers for archive: faraday, wget; using faraday
```
Another failure happens later on when the faraday_middleware gem isn't installed:
```
  no such file to load -- faraday_middleware
```

Resolved this by confining the faraday provider so it only
gets used when the faraday_middleware gem is installed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ group :test do
   gem 'rspec-puppet-facts',                                         :require => false
   gem 'rspec',                                                      :require => false
   gem 'puppet-blacksmith',                                          :require => false
-  gem 'rubocop',                                                    :require => false
+  gem 'rubocop', '0.35.0',                                          :require => false
   gem 'puppetlabs_spec_helper',                                     :require => false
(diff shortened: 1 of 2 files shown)


igalic:
> Error: The `Style/TrailingComma` cop no longer exists. Please use `Style/TrailingCommaInLiteral` and/or `Style/TrailingCommaInArguments` instead.

we need a modulesync here

jyaworski:
@epienbroek please make this change, in the meantime.

https://github.com/voxpupuli/puppet-collectd/blob/master/Gemfile#L26

epienbroek:
I just updated this PR and the CI checks are all green now

jyaworski:
Thanks. Could you squash down into one commit?

epienbroek:
Done

jyaworski:
Thanks!

Issue #109, opened 2016-01-09
params gem provider logic fails with PE puppet 2015.03


logicminds:
Just downloaded PE 2015.03 and the params code that determines which gem provider to use is not working correctly because $::aio_agent_version does not exist. Maybe this should be a case statement comparing all the versions since PE puppet no longer contains 'puppet enterprise' in the newer versions

https://github.com/voxpupuli/puppet-archive/blob/master/manifests/params.pp#L22

similar to #106 but slightly different

logicminds:
```puppet
if $::puppetversion =~ /Puppet Enterprise/ and $::osfamily != 'Windows' {
    $gem_provider = 'pe_gem'
  } elsif $::aio_agent_version {
    $gem_provider = 'puppet_gem'
  } else {
    if versioncmp($::puppetversion, '4.0.0') >= 0 {
      $gem_provider = 'puppet_gem'
    } else {
      $gem_provider = 'gem'
    }
  }

```
The above code seems to work for me.  I am not sure when PE puppet stoppped displaying puppet enterprise in the puppetversion though.

xaniasd:
https://tickets.puppetlabs.com/browse/FACT-1280

logicminds:
Also this issue was recently merged which might change some of the code above.

https://github.com/puppetlabs/puppetlabs-pe_gem/pull/10

aerostitch:
Hi,

The latest changes brought by #117 and #121 should bring what you asked for here. Can you confirm please?
If so I think this can be closed right?

Thanks
Joseph (JH) ;)

jyaworski:
Closing due to recently-merged PRs. Please re-open if this is still an issue.

PR #117, opened 2016-02-17
Fixing tests failures on non-defined facts

aerostitch:
Hi,

Currently the tests are failing on testing of undefined facts when compiling with STRICT_VARIABLES:
```
Undefined variable "::aio_agent_version"; Undefined variable "aio_agent_version" at /home/travis/build/voxpupuli/puppet-archive/spec/fixtures/modules/archive/manifests/params.pp:24 on node testing-worker-linux-docker-bc362139-3367-linux-14.prod.travis-ci.org
```
and:
```
Undefined variable "::archive_windir"; Undefined variable "archive_windir" at /home/travis/build/voxpupuli/puppet-archive/spec/fixtures/modules/archive/manifests/params.pp:13 on node testing-worker-linux-docker-bc362139-3367-linux-14.prod.travis-ci.org
```

Example: https://travis-ci.org/voxpupuli/puppet-archive/jobs/109740076

To fix this we can use getvar. This PR fixes that.

Thanks
Joseph

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,17 @@ class { 'archive':
 }
 
 ```
+
+Note: If you are writing a dependent module that include specs in it, you will
+need to set the puppetversion fact in your puppet-rspec tests. You can do that
+by adding it to the default facts of your spec/spec_helper.rb. Example:
+
+```
+RSpec.configure do |c|
+  c.default_facts = { :puppetversion => Puppet.version }
+end
+```
+
 ## Reference
 
 ### Classes
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,14 +1,6 @@
 # archive parameters
 class archive::params {
   case $::osfamily {
-    default: {
-      $path  = '/opt/staging'
-      $owner = '0'
-      $group = '0'
-      $mode  = '0640'
-      $seven_zip_name = undef
-      $seven_zip_provider = undef
-    }
     'Windows': {
       $path               = $::archive_windir
       $owner              = 'S-1-5-32-544' # Adminstrators
@@ -16,14 +8,24 @@ class archive::params {
       $mode               = '0640'
       $seven_zip_name     = '7zip'
       $seven_zip_provider = 'chocolatey'
+      if versioncmp($::puppetversion, '4.3.1') >= 0 {
+        $gem_provider = 'pe_gem'
+      } else {
+        $gem_provider = 'gem'
+      }
     }
-  }
-
-  if $::puppetversion =~ /Puppet Enterprise/ and $::osfamily != 'Windows' {
-    $gem_provider = 'pe_gem'
-  } elsif $::aio_agent_version {
-    $gem_provider = 'puppet_gem'
-  } else {
-    $gem_provider = 'gem'
+    default: {
+      $path  = '/opt/staging'
+      $owner = '0'
+      $group = '0'
+      $mode  = '0640'
+      $seven_zip_name = undef
+      $seven_zip_provider = undef
+      if $::puppetversion =~ /Puppet Enterprise/ or versioncmp($::puppetversion, '4.0.0') >= 0 {
+        $gem_provider = 'pe_gem'
+      } else {
+        $gem_provider = 'gem'
+      }
+    }
   }
 }
--- a/spec/classes/archive_spec.rb
+++ b/spec/classes/archive_spec.rb
@@ -17,8 +17,16 @@ describe 'archive' do
     it { is_expected.to_not contain_package('7zip') }
   end
 
+  context 'RHEL Puppet 4.0.0 opensource' do
+    let(:facts) { { :osfamily => 'RedHat', :puppetversion => '4.0.0' } }
+
+    it { is_expected.to contain_package('faraday').with_provider('pe_gem') }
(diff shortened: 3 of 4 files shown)


jyaworski:
I would rather use versioncmp and puppetversion as it's more idiomatic. We can sidestep aio_agent_version.

aerostitch:
Ok. I just saw #109, so I integrated the part that checks the version from @logicminds and also dropped the aio_agent_version fact call.

But it still needs to have the archive_windir fact checked for undef.

aerostitch:
Note that we don't take in account the specific case Antoine explained in https://tickets.puppetlabs.com/browse/FACT-1280 (having puppet running with 4.1 and system/os ruby when installed using the OS package on Fedora).
But we were not taking it in account in the first place either, and I'm not sure if it is possible at this stage.

logicminds:
To make things more complex the gem provider for windows was broken up until 4.3.1. I back ported a fix and made the pe_gem provider work on windows. So if your using the latest pe_gem provider code we can use the following code now.  This would require the usage of the latest pe_gem.

Note: this is untested, I think it would work

```puppet
if $::puppetversion =~ /Puppet Enterprise/  {
   # this is for puppet enterprise installs only, and only works on PE 3.x
    $gem_provider = 'pe_gem'  # now works on windows
} else {  
    if versioncmp($::puppetversion, '4.0.0') >= 0 {
      $gem_provider = 'puppet_gem'
    } else {
      $gem_provider = 'gem'
    }
}
```

aerostitch:
Thanks, in this case, let me just add a specific case for windows with puppet 4.3.1! ;)

logicminds:
Well technically you could have a special case for windows to always use pe_gem (even if not using pe) and then use puppet_gem if later than 4.3.0.  Although this might confuse people.

aerostitch:
Yep, that's the change I just pushed while you were hitting "comment"! ;)

aerostitch:
Updated the commit taking your advice in account. :)
Thanks

manifests/params.pp
```diff
@@ -1,6 +1,19 @@
 # archive parameters
 class archive::params {
   case $::osfamily {
+    'Windows': {
+      $path               = getvar('::archive_windir')
+      $owner              = 'S-1-5-32-544' # Adminstrators
+      $group              = 'S-1-5-18'     # SYSTEM
```
jyaworski: What is this sorcery?

aerostitch: I don't follow. What sorcery?

jyaworski: Sorry, I was being facetious and referring to that S-1-5-32-544 'sorcery.'

manifests/params.pp
```diff
@@ -8,22 +21,11 @@
...
-      $mode               = '0640'
-      $seven_zip_name     = '7zip'
-      $seven_zip_provider = 'chocolatey'
-    }
-  }
-
-  if $::puppetversion =~ /Puppet Enterprise/ and $::osfamily != 'Windows' {
-    $gem_provider = 'pe_gem'
-  } elsif $::aio_agent_version {
-    $gem_provider = 'puppet_gem'
```
logicminds: The puppet_gem should be the default for non-windows and puppet > 4.0

aerostitch: It is, check the lines added (green section) :)

In the default case:
```
 if $::puppetversion =~ /Puppet Enterprise/ or versioncmp($::puppetversion, '4.0.0') >= 0{
```

manifests/params.pp
aerostitch: Looool!
That's the "Administrators" users according to the Windows convention: https://support.microsoft.com/en-us/kb/243330
No wonder why it's been years I haven't touched a Windows box! :)

aerostitch:
Don't merge yet, I still need to do an update to get rid of the getvar usage.

jyaworski:
Please add [WIP] to the PR title.

jyaworski:
I just ran into this on [puppet-rundeck](https://travis-ci.org/voxpupuli/puppet-rundeck/jobs/110242268).

aerostitch:
Hi guys,

I updated the specs and params to fix the issue, cover the tests of different providers with newer versions of puppet and avoid using the getvar method for the archive_windir.
I also added a rspec test for the archive_windir fact at the same time.

Can you review and tell me if those changes are ok for you please?

Thanks
Joseph

aerostitch:
Note: I just added a note in the readme to help people that write modules depending on this module to avoid searching an issue for nothing when it comes to errors with the required fact puppetversion (not present in puppet-rspec by default). I think it could save some time to anyone integrating this module as a dependendcy of another one.

README.md
```diff
@@ -87,6 +87,17 @@ class { 'archive':
...
 
 ```
+
+Note: If you are writing a dependent module that include specs in it, you will
+need to set the puppetversion fact in your puppet-rspec tests. You can do that
+by adding it to the default facts of your spec/spec_helper.rb. Example:
+
+```
+RSpec.configure do |c|
+  c.default_facts = { :puppetversion => `puppet --version` }
```
jyaworski: You could use Puppet.version here.

jyaworski:
Besides my one comment on the README, this LGTM.

aerostitch:
Excellent advise on the Puppet.version.
Updated the PR accordingly.
Thanks! :)

PR #119, opened 2016-02-17
Adding more Puppet versions to Travis

aerostitch:
Hi,

There have been significant changes in several versions of Puppet 3.
It's a good practice to test the different versions. Also adding the support of the future parser that brings also significant changes.

For example, when I build a dependent package with Travis on puppet 3.5 to 3.8 I get:
```
     Puppet::Error:
       Undefined variable "::puppetversion"; Undefined variable "puppetversion" at /home/travis/build/tubemogul/puppet-aerospike/spec/fixtures/modules/archive/manifests/params.pp:22 on node testing-worker-linux-docker-94746611-3371-linux-6.prod.travis-ci.org
```
I'm looking for a fix for the generated error and I'll add it to this PR.
For now, if you could just tell me if you'd be ok to add those environments to the travis builds, it would be awesome.

Thanks! :)
Joseph

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,10 @@ matrix:
   - rvm: 2.1.6
     env: PUPPET_VERSION="~> 3.0" STRICT_VARIABLES="yes" CHECK=test
   - rvm: 2.1.6
+    env: PUPPET_VERSION="~> 3.5" STRICT_VARIABLES="yes" CHECK=test
+  - rvm: 2.1.6
+    env: PUPPET_VERSION="~> 3.8" STRICT_VARIABLES="yes" FUTURE_PARSER=yes CHECK=test
+  - rvm: 2.1.6
     env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test
   - rvm: 2.2.3
     env: PUPPET_VERSION="~> 4.0" STRICT_VARIABLES="yes" CHECK=test


jyaworski:
Travis is managed by modulesync, and should not be edited directly:

https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.travis.yml

Please raise PRs about this on that repository.

aerostitch:
Oh thanks.
Actually that means we need a config of modulesync for this repo.

jyaworski:
This module is [already managed by modulesync](https://github.com/voxpupuli/modulesync_config/blob/master/managed_modules.yml#L3).

aerostitch:
Hi,

Created the PR on the modulesync config. Not sure what's the process then and if we need to keep this PR open.

Thanks
Joseph

aerostitch:
Hmm... It seems that the issue I have might be due to the fact that the puppetversion fact has been removed from the core facts and re-added later:
https://tickets.puppetlabs.com/browse/FACT-1277
https://tickets.puppetlabs.com/browse/PUP-4359

@jyaworski would you be ok if I push the puppetversion fact call inside a getvar in params.pp to bypass this issue?

aerostitch:
Closing this as won't fix.
I'll wait for the pull request that fixes the tests to be integrated to be able to move on in the debugging of my own tests failures.

PR #121, opened 2016-02-20
Fixing the provider name bug introduced in #117

aerostitch:
Hi,

I introduced a bug in #117 : I mixed up the name of the embedded gem provider for puppet >= 4. I put pe_gem instead of the new 'puppet_gem' http://docs.puppetlabs.com/puppet/4.0/reference/type.html#package-provider-puppet_gem.
pe_gem is the one coming from https://github.com/puppetlabs/puppetlabs-pe_gem which is deprecated in puppet 4.

This one fixes this error and fixes the related tests.

Sorry for that. :/
Thanks
Joseph


aerostitch:
Thanks! ;)

PR #122, opened 2016-02-21
Fixes GH-77

### LINKED ISSUE #77: allow source to accept absolute file paths
I have been noticing that I constantly have to convert /tmp/some/file.txt to file:///tmp/some/file.txt and would rather just have the type code do this by munging the value. 

Any objection to adding support for the source parameter to accept absolute file paths?

jyaworski:
Fixes #77

### DIFF
--- a/lib/puppet_x/bodeco/util.rb
+++ b/lib/puppet_x/bodeco/util.rb
@@ -4,7 +4,7 @@ module PuppetX
   module Bodeco
     module Util
       def self.download(url, filepath, options = {})
-        uri = URI(url)
+        uri = Puppet::Util.path_to_uri(url) unless url.is_a?(URI)
         @connection = PuppetX::Bodeco.const_get(uri.scheme.upcase).new("#{uri.scheme}://#{uri.host}:#{uri.port}", options)
         @connection.download(uri, filepath)
       end
@@ -96,7 +96,7 @@ module PuppetX
       end
 
       def download(uri, file_path)
-        FileUtils.copy(uri.path, file_path)
+        FileUtils.copy(Puppet::Util.uri_to_path(uri), file_path)
       end
     end
   end


lib/puppet_x/bodeco/util.rb
```diff
@@ -4,7 +4,7 @@ module PuppetX
   module Bodeco
     module Util
       def self.download(url, filepath, options = {})
-        uri = URI(url)
+        uri = Puppet::Util.path_to_uri(uri) unless uri.is_a?(URI)
```
nanliu: The spec is failing here because you should use url instead of uri, otherwise url is an undeclared variable:
Puppet::Util.path_to_uri(url)
instead of:
Puppet::Util.path_to_uri(uri)

jyaworski: I need a new font. Fixed.

lib/puppet_x/bodeco/util.rb
```diff
@@ -4,7 +4,7 @@ module PuppetX
   module Bodeco
     module Util
       def self.download(url, filepath, options = {})
-        uri = URI(url)
+        uri = Puppet::Util.path_to_uri(url) unless url.is_a?(URI)
```
nanliu: The unless will always be false since we removed the URI(url) line above. In this case if the Puppet::Util.path_to_uri is smart maybe just remove the unless. Need to double check their code to be sure.

nanliu:
We should abandon this route. I've got a user report the curl/wget provider doesn't understand file:/// and attempts to curl/wget a local file. I'll try to write a patch to fix both issues.

nanliu:
See #123 for alternative which address the local file bug as well.

nanliu:
Closed, fixed by #123.

Issue #78, opened 2015-10-22
file urls do not work on windows


logicminds:
Because windows has to be different...

The URI ruby library does not work as expected on windows because of the fact there is no root drive.  So files fail to download because the urls are interpreted incorrectly. 

Example:

```shell
irb(main):006:0> URI('file://d:/OSS/one/two/three.zip').path
=> "/OSS/one/two/three.zip"

```

And adding another slash produces another issue

```shell
irb(main):007:0> URI('file:///d:/OSS/one/two/three.zip').path
=> "/d:/OSS/one/two/three.zip"
```

This should be d:/OSS/one/two/three.zip

logicminds:
I have this working on my own fork with the following:


```ruby
class FILE
      def initialize(_url, _options)
      end

      def normalize_path(uri)
        # URI for windows does not work as expected and returns a path without the root drive
        # this attempts to fix that
        if /^file:\/{2}([a-z,A-Z]?)/.match(uri.to_s)   # file://c:/one/two/three.zip
          uri.to_s.gsub(/^file:\/{2}([a-z,A-Z])?/, "#{$1}:")
        else
          uri.path
        end
      end
      
      def download(uri, file_path)
        src_path = normalize_path(uri)
        FileUtils.copy(src_path, file_path)
      end
    end
```

If this is acceptable I can make a PR.

nanliu:
file:///d:/OSS/one/two/three.zip is technically correct, because I think file://d:/OSS/one/two/three.zip, would be host = d:, path = OSS, and this patch is implementing the wrong behavior (puppet:/// and not puppet:// because puppet://{server}/). So maybe implementing #77 is the better approach, because this is not intuitive for Windows.

PR #123, opened 2016-02-24
Fix several issues related to local files

nanliu:
* Fix local files for non faraday providers
* Fix GH-77 allows local files to be specified without using file:///
* Fix GH-78 allow local file:///c:/... on windows

### DIFF
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -117,7 +117,16 @@ Puppet::Type.type(:archive).provide(:ruby) do
     temppath = tempfile.path
     tempfile.close!
 
-    download(temppath)
+    case resource[:source]
+    when /^(http|ftp)/
+      download(temppath)
+    when /^file/
+      uri = URI(resource[:source])
+      FileUtils.copy(Puppet::Util.uri_to_path(uri), temppath)
+    else
+      fail(Puppet::Error, "Source file: #{resource[:source]} does not exists.") unless File.exist?(resource[:source])
+      FileUtils.copy(resource[:source], temppath)
+    end
 
     # conditionally verify checksum:
     if resource[:checksum_verify] == :true && resource[:checksum_type] != :none
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -103,7 +103,7 @@ Puppet::Type.newtype(:archive) do
   newparam(:source) do
     desc 'archive file source, supports http|https|ftp|file uri.'
     validate do |value|
-      unless value =~ URI.regexp(%w(http https file ftp))
+      unless value =~ URI.regexp(%w(http https file ftp)) || Puppet::Util.absolute_path?(value)
         fail ArgumentError, "invalid source url: #{value}"
       end
     end
--- a/spec/unit/puppet/type/archive_spec.rb
+++ b/spec/unit/puppet/type/archive_spec.rb
@@ -26,16 +26,34 @@ describe Puppet::Type.type(:archive) do
     end.to raise_error(Puppet::Error, /archive path must be absolute: /)
   end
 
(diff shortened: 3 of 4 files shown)


nanliu:
WIP, do not merge yet.

lib/puppet/type/archive.rb
```diff
@@ -103,7 +103,7 @@ def should_to_s(value)
   newparam(:source) do
     desc 'archive file source, supports http|https|ftp|file uri.'
     validate do |value|
-      unless value =~ URI.regexp(%w(http https file ftp))
+      unless value =~ URI.regexp(%w(http https file ftp)) || value =~ %r{^([A-Z]:|\/)}
```
jyaworski: There's a Puppet::Util.Absolute_Path? or something.

lib/puppet/provider/archive/ruby.rb
```diff
@@ -117,7 +117,16 @@ def transfer_download(archive_filepath)
     temppath = tempfile.path
     tempfile.close!
 
-    download(temppath)
+    case resource[:source]
+    when /^(http|ftp)/
+      download(temppath)
+    when /^file/
+      uri = URI(resource[:source])
+      FileUtils.copy(uri.path, temppath)
```
jyaworski: Does this solve the Windows problem?

lib/puppet/type/archive.rb
nanliu: yes, do you know how stub os correctly to run rspec for Puppet::Util.absolute_path? otherwise spec tests will fail for Windows test cases.

lib/puppet/provider/archive/ruby.rb
nanliu: Which issue? it does not address #78.

lib/puppet/type/archive.rb
jyaworski: It looks like the function has support for Windows already. Do you mean something else?

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util.rb#L237

nanliu: I meant I run into problems testing windows filepath in rspec if I use Puppet::Util.absolute_path?

```puppet
  1) Puppet::Type::Archive verify resoource[:source] is valid source
     Failure/Error: expect do
       expected no Exception, got #<Puppet::ResourceError: Parameter source failed on Archive[/tmp/example.zip]: invalid source url: D:/example.zip> with backtrace:
         # ./.bundle/gems/puppet-3.8.4/lib/puppet/util/errors.rb:106:in `fail'
         # ./lib/puppet/type/archive.rb:107:in `block (3 levels) in <top (required)>'
         # ./.bundle/gems/puppet-3.8.4/lib/puppet/parameter.rb:465:in `validate'
         # ./.bundle/gems/puppet-3.8.4/lib/puppet/parameter.rb:500:in `value='
         # ./.bundle/gems/puppet-3.8.4/lib/puppet/type.rb:654:in `[]='
         # ./spec/unit/puppet/type/archive_spec.rb:35:in `block (3 levels) in <top (required)>'
         # ./spec/unit/puppet/type/archive_spec.rb:30:in `block (2 levels) in <top (required)>'
     # ./spec/unit/puppet/type/archive_spec.rb:30:in `block (2 levels) in <top (required)>'
```

The spec behavior is different depending on where you run it. On Mac/Linux it expects / path, and Windows [A-Z]: path. I'm looking for a proper way to stub platform detection for this method, otherwise rake spec will fail.

I suppose as last resort I can get rid of the test cases for the file path and just trust it does the right thing.

jyaworski: In my opinion, trusting the mothership to do the right thing is a good idea until we specifically find out that we need to work around it.

nanliu: Expect to be resolved in latest PR.

Issue #55, opened 2015-05-09
Out of memory while downloading large files


pall-valmundsson:
While trying to download a large .zip file (1.3GB) in a test VM with 512MB of RAM the only output I got from the puppet run was `Killed`. I increased the VM RAM incrementally and got `Error: Could not run: failed to allocate memory` until I had increased the RAM to 2GB, i.e. the file could be stored in RAM as a whole until flushed to disk.

Is this a known unfixable issue? Or an issue with the provider implementation? Or is there some workaround that can be applied?

nibalizer:
Are you pulling the file over http or ftp?

pall-valmundsson:
Ah, sorry I left that out. This is a HTTP Basic Auth pull.

nibalizer:
Looking a bit at the code, this looks like an issue in the faraday library. However I dont see an issue on the faraday bug tracker for it. I'll try to find some time to bang on it today.

pall-valmundsson:
Thanks for looking into this. Other potentially important information.
Puppet 3.7.5
CentOS 6
and therefore Ruby 1.8.7

nibalizer:
I am able to reproduce the problem at least.

nanliu:
In this case might as well swap out faraday with net/http and eliminate a dependency. I'll take a crack at it over next few days.

nibalizer:
@nanliu do you think its in faraday? Is the typical pattern in puppet to hold in ram before file flushing? I'm just not experienced enough to know.

nanliu:
I thought this was one of things I verified early on, but clearly missed it when evaluating faraday. Per their roadmap: https://github.com/lostisland/faraday/wiki/Roadmap and open PR: https://github.com/lostisland/faraday/pull/156, I don't think we are able to do streaming with faraday. net/http will require some code to handle http errorcodes and redirect, but will support streaming.

jyaworski:
Since it looks like the Faraday PR is stalled, should we move to net/http or similar?

nanliu:
Yes, we should switch to net/http as default provider.

Issue #118, opened 2016-02-17
embed gems and install locally


logicminds:
Not sure if anybody else can appreciate doing this, but my current situation limits me from installing anything directly off the internet.  So to make things easier I ended up putting the gems inside the module and installing locally. 

If this kind of addition is welcomed, I'll make a PR for it. 

But it looks something like:

```puppet
  $archive::params::path               = 'C:/Windows/Temp'  # for context only
  $faraday_source            = "${archive::params::path}/faraday-0.9.1.gem"
  $faraday_middleware_source = "${archive::params::path}/faraday_middleware-0.10.0.gem"
  $multipart_post_source     = "${archive::params::path}/multipart-post-2.0.0.gem"

  file{$faraday_source:
    ensure => present,
    source => 'puppet:///modules/archive/faraday-0.9.1.gem',
    before => Package['faraday']
  }
  file{$faraday_middleware_source:
    ensure => present,
    source => 'puppet:///modules/archive/faraday_middleware-0.10.0.gem',
    before => Package['faraday_middleware']
  }
  file{$multipart_post_source:
    ensure => present,
    source => 'puppet:///modules/archive/multipart-post-2.0.0.gem',
    before => Package['multipart-post']
  }
  package{ 'multipart-post':
    ensure   => present,
    provider => $archive::params::gem_provider,
    install_options => '--local',
    source   => $multipart_post_source,
  }
  package { 'faraday':
    ensure   => present,
    provider => $archive::params::gem_provider,
    install_options => '--local',
    source   => $faraday_source,
    require  => Package['multipart-post'],
  }
  package { 'faraday_middleware':
    ensure   => present,
    provider => $archive::params::gem_provider,
    install_options => '--local',
    source   => $faraday_middleware_source,
    require  => Package['faraday'],
  }


```

jyaworski:
Isn't this the kind of thing that packaging as RPMs solves?

logicminds:
RPMs only work on redhat flavored linux. :(

jyaworski:
I've used [fpm](https://github.com/jordansissel/fpm/wiki/ConvertingGems) to build Debian packages as well.

logicminds:
Also using Windows and AIX as well.  So thats three platforms I have to build packages for, which is why I just bundled them into the module.

jyaworski:
In that case, I'm wondering if that is out of the scope of puppet-archive and should belong in a profile.

@nanliu @daenney

[...]

PR #124, opened 2016-02-25
GH-55 use net::http to stream files

nanliu:
* Fix GH-55 large file memory issue.
* Remove puppet agent dependency on Faraday gem.
* make ruby default provider for windows.
* migrate functions away from faraday.
* change target release to 0.5.0


nanliu:
WIP, I'm not sure level of effort to remove faraday from the existing functions and difficult to test without those services. Also still missing following features:

* follow http redirect
* http cookies

Need more testing for ruby provider handling of:
* proxy
* username/password

nanliu:
@logicminds, would you be able interested to test this in your Windows environment and see if it resolves your issue so you don't need #118 anymore?

jyaworski:
Are we just killing Faraday? Net::Http is likely better.

nanliu:
Yeah, trying to remove the dependency since there was no way to fix #55 with faraday. The tricky part is dealing with the three functions https://github.com/voxpupuli/puppet-archive/blob/master/lib/puppet/parser/functions/artifactory_sha1.rb#L16-L22. Need some volunteers for testing those changes.

dandunckelman:
I came up against this yesterday.  Impeccable timing.  If I get a chance, I'll see if I can help out.  Tied up right now.

jyaworski:
I can actually test artifactory, but I need to see if it's been rolled out yet...

nanliu:
@jyaworski I tested the new functions against jfrog's repo and added some spec tests. I couldn't find an example gocd md5 file and wrote it test based from memory. Anyhow, additional testing is appreciated.

jyaworski:
@nanliu this doesn't appear to follow redirects, and it also looks like it's hardcoded to http in the art_url in the define.

```puppet
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: unexpected token at '<html>
<head><title>400 The plain HTTP request was sent to HTTPS port</title></head>
<body bgcolor="white">
<center><h1>400 Bad Request</h1></center>
<center>The plain HTTP request was sent to HTTPS port</center>
<hr><center>nginx</center>
</body>
</html>
' at /etc/puppet/environments/archive_artifactory/modules/archive/manifests/artifactory.pp:38 on node domain.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

```puppet
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: unexpected token at '<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
' at /etc/puppet/environments/archive_artifactory/modules/archive/manifests/artifactory.pp
```

jyaworski:
Following up from IRC, maybe we could drop the port parameter and pass a URI to be parsed. That way net::http can do its own thing.

nanliu:
@jyaworski, both fixed, it should follow redirect for content, and use the correct url (instead of hard coded http in the manifest for checksum url).

jyaworski:
That works. Maybe we should add more docs; it wasn't immediately clear that URL was the whole URL to the entire file.

Or maybe we should just deprecate port, and keep the base url (with :port/artifactory) and filepath separate like it is currently.

nanliu:
I updated the docs significantly. After looking at archive::nexus, I can see how url as a parameter might be confusing. I didn't author nexus, so I have to dig around a bit to understand the parameters. If you have an alternate suggestion, let me know.

jyaworski:
This LGTM now. I would say this is ready to be merged.

jyaworski:
After looking over this, I'm willing to merge it now.

PR #138, opened 2016-03-14
open file in binary mode when writing downloaded content

damoxc:
On Windows writing data to files in text mode will modify the resulting content
of the file, which causes the downloaded file to break. Add the binary flag to
the file mode to ensure there is no character conversion and the data is written
as-is.


igalic:
that seems like a very sensible thing to do
i wish there was a way to test for this…

jyaworski:
I'm going to merge. This seems sensible.

Issue #114, opened 2016-01-29
support stripping of directories


igalic:
many compression tools support stripping of a number of directories, we should too

jyaworski:
Can you give an example?

eljohn1014:
I am looking for this feature as well. can you add in an option to do strip-components=<num> so that we can use the --strip-components=<num> flag in tar

nanliu:
That would be very specific to tar. Does the current extract_flags option suffice?

igalic:
it would be nice to be able to have agnostic of the underlying archive format's support for the flag.

jyaworski:
Would [dirname](https://github.com/puppetlabs/puppetlabs-stdlib#dirname) suffice?

dhoppe:
I was able to solve this by using the parameter ```extract_command``` as you can see at the [kafka](https://github.com/voxpupuli/puppet-kafka/pull/50) module.

jyaworski:
Given @dhoppe's great answer, I'm going to mark this as closed.

PR #139, opened 2016-03-15
GH-114 Document extract customization options

nanliu:
* Provide example for stripping directories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
    * [Example](#usage-example)
    * [File permission](#file-permission)
    * [Network files](#network-files)
+   * [Extract customization](#extract-customization)
    * [S3 Bucket](#s3-bucket)
 5. [Reference](#reference)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -211,6 +211,23 @@ archive { '/nfs/repo/software.zip':
 }
 ```
 
+### Extract Customization
+
+The `extract_flag` or `extract_command` parameters can be used to override the
```
rnelson0: Can you give an example of extract_flag as well, or is it a simple alias? If the latter maybe a trailing comment on line 224 saying '# you can also use extract_flag" or similar.

nanliu: Example added for the simple extract_flag usage. The one other use case is if you need a different flag for 7z on windows v.s. another command in linux which can be done as a hash of flags, but in practice I haven't run into it and I'm not sure if it's a feature that we want to promote.

Issue #140, opened 2016-03-16
$aws_cli_install defaults to true for non Amazon instances


danifr:
getvar('::ec2_metadata') seems to return true if fact does not exist.

https://github.com/voxpupuli/puppet-archive/blob/master/manifests/params.pp#L27

jyaworski:
Can we try changing getvar to $::?

danifr:
This seems to work:
if getvar('::ec2_metadata') != '' and $::operatingsystem != 'Amazon' {

But I can test you suggestion too. One sec.

danifr:
Arggg, I cannot test your change right now. I need to recreate the VM....

Anyway... I'll submit a PR with my change and we can discuss in there if we want to use your suggestion or stick with mine.

nanliu:
The ec2_* facts only exist in aws environment so we can't use $:: alone.

danifr:
Closing https://github.com/voxpupuli/puppet-archive/pull/141

PR #141, opened 2016-03-16
Fix aws_cli_install condition

danifr:
https://github.com/voxpupuli/puppet-archive/issues/140

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -24,7 +24,7 @@ class archive::params {
   }
 
   # Amazon Linux already have aws cli installed:
-  if getvar('::ec2_metadata') and $::operatingsystem != 'Amazon' {
+  if getvar('::ec2_metadata') != '' and $::operatingsystem != 'Amazon' {
     $aws_cli_install = true
   } else {
     $aws_cli_install = false


manifests/params.pp
```diff
@@ -24,7 +24,7 @@
   }
 
   # Amazon Linux already have aws cli installed:
-  if getvar('::ec2_metadata') and $::operatingsystem != 'Amazon' {
+  if getvar('::ec2_metadata') != '' and $::operatingsystem != 'Amazon' {
```
jyaworski: I am skeptical of doing this under puppet 4 (or in general, but specifically under puppet 4). Does

```puppet
  if $::ec2_metadata and $::operatingsystem != 'Amazon' { 
```

work?

danifr: Sorry I messed it up, there are no changes needed.
This code is working just fine. `getvar('::ec2_metadata')` returns the content of the ec2_metadata fact as it is expected.

Anyway I would suggest changing this condition. I'm not using an ec2 VM but for some reason the fact ec2_metadata is not empty. Because of this, it returns true and then the module tries to install a bunch of packages (failing if unzip is not installed....).

Would it be a problem changing the default value of `$aws_cli_install` to false, and letting the user changing it if interested?

jyaworski: What version of facter are you running? I don't see this in a newer facter.

```shell
[root@puppet2 ~]# facter -p facterversion ec2_metadata
ec2_metadata => 
facterversion => 3.1.4
```
```shell
[root@puppetdb1 ~]# facter -p facterversion ec2_metadata
ec2_metadata => nil
facterversion => 2.4.6
```

jyaworski: Tagging @nanliu. Is using getvar a way to also get around strict_variables? It says it's the same as $::, so it seems like a cop-out.

igalic:
`Evaluation Error: Unknown variable: '::operatingsystem'` ← now that's just ridiculous.

danifr:
I'm running facterversion => 2.4.6

This might be something related to the configuration applied in my organization.
I'm not using ec2 VMs (we actually use OpenStack) but for some reason ec2_metadata returns data.
```
$ facter -p facterversion ec2_metadata
Already evaluated rest at /usr/local/share/gems/gems/facter-2.4.6/lib/facter/gce.rb:4, reevaluating anyways
ec2_metadata => {"ami-id"=>"ami-000005c3", "ami-launch-index"=>"0", "ami-manifest-path"=>"FIXME", "block-device-mapping"=>{"ami"=>"vda", "root"=>"/dev/vda"}, "hostname"=>"cci-rundeck-qa-01", "instance-action"=>"none", "instance-id"=>"i-000a4027", "instance-type"=>"m1.small", "kernel-id"=>"None", "local-hostname"=>"cci-rundeck-qa-01", "local-ipv4"=>"111.111.222.111", "placement"=>{"availability-zone"=>"zone-a"}, "public-hostname"=>"cci-rundeck-qa-01", "public-ipv4"=>nil, "ramdisk-id"=>"None", "reservation-id"=>"r-0q8h85ky", "security-groups"=>"default"}
```
This is the reason why the condition does not work for my case.

jyaworski:
```rspec
         expected: (["http://home.lan/example.zip", "-O", String, "--max-redirect=5", "--header=\"Cookie: \"foo\""])

              got: (["http://home.lan/example.zip", "-O", "/tmp/example.zip", "--max-redirect=5", "--header=\"Cookie: foo\""])
```

Please change the spec test.

danifr:
@jyaworski I guess your last comment was meant for the other PR.
I'm closing this one and opening a separate issue to discuss whether makes sense or not to have this condition.

Issue #142, opened 2016-03-16
Cannot download using wget


randradas:
When wget is used my download fail. It seems that the server is not receiving a proper cookie header. I checked my puppet agent log and I saw this:

> Debug: Executing '/usr/bin/wget http://download.oracle.com/otn-pub/java/jdk/7u80-b15/jdk-7u80-linux-x64.tar.gz -O /tmp/jdk-7u80-linux-x64.tar.gz_20160316-9551-fy9sql --max-redirect=5 --header="Cookie: "oraclelicense=accept-securebackup-cookie"'

If I run this manually in my shell, I got a wrong download. As you can see cookie header double quotes are wrong.

I tried to change **wget.rb** from this (three double quotes on Cookie):
`params += optional_switch(resource[:cookie], ['--header="Cookie: "%s"'])`

To this (two double quotes on Cookie):
`params += optional_switch(resource[:cookie], ['--header="Cookie: %s"'])`

web server is still answering like it is receiving a wrong cookie but in this case If I run the command printed by the log in my shell manually then it works.

igalic:
@randradas would you like to provide us with a patch for this?

randradas:
yes!, I would. I've just push it on pull requests.

nanliu:
Odd the original spec 4c469d3 had the extra double quote, wonder how it was working originally. Is this fixed by #143?

randradas:
as far as I know and according with wget man page, only double quotes before '--header=' and the final one are needed. https://www.gnu.org/software/wget/manual/wget.html. #143 removes this extra double quote that should not be there according with wget documentation.

nanliu:
When I'm testing with the following resource, I only get an html page:

```puppet
archive { '/tmp/jdk-7u80-linux-x64.tar.gz':
  source   => 'http://download.oracle.com/otn-pub/java/jdk/7u80-b15/jdk-7u80-linux-x64.tar.gz',
  cookie   => 'oraclelicense=accept-securebackup-cookie',
  provider => 'wget',
}
```

```bash
$ file /tmp/jdk-7u80-linux-x64.tar.gz
/tmp/jdk-7u80-linux-x64.tar.gz: HTML document text
$ head /tmp/jdk-7u80-linux-x64.tar.gz
<html>
<head>
<title>Unauthorized Request</title>
<META NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">
[...]

PR #147, opened 2016-03-17
GH-142 Fix wget cookies option

nanliu:


nanliu:
Most likely breaking spec tests, provided as work around. I'll resolve spec tests later.

lib/puppet/provider/archive/wget.rb
```diff
@@ -14,6 +14,9 @@ def download(filepath)
     params += optional_switch(resource[:cookie], ['--header="Cookie: %s"'])
     params += optional_switch(resource[:proxy_server], ["--#{resource[:proxy_type]}_proxy=#{resource[:proxy_server]}"])
 
-    wget(params)
+    # NOTE:
+    # Do NOT use wget(params) until https://tickets.puppetlabs.com/browse/PUP-6066 is resolved.
+    command = "wget #{params.join(' ')}"
```
jyaworski: This makes me sad.

igalic: huh? it doesn't just use Shellwords?

PR #155, opened 2016-04-19
Problem with URL generated by function assemble_nexus_url when version contains a plus sign ("+")

thiagomarinho:
I had a problem when I tried to download an artifact from local nexus which version contains a "+" symbol. Nexus recognizes it as a "space", and I couldn't download the artifact. So I added the CGI.escape method wrapping the parameter values to build a valid url to nexus:

```
query_string = params.to_a.map { |x| "#{x[0]}=#{CGI.escape(x[1])}" }.join('&')
```

As only in archive::nexus the module is responsible to build the URL, I think this is the right way to correct it.

### DIFF
--- a/lib/puppet/parser/functions/assemble_nexus_url.rb
+++ b/lib/puppet/parser/functions/assemble_nexus_url.rb
@@ -1,10 +1,12 @@
+require 'cgi'
+
 module Puppet::Parser::Functions
   newfunction(:assemble_nexus_url, :type => :rvalue) do |args|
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hey @thiagomarinho, thanks for the contribution! This looks fine, just a little rubocop blaming. Could you take a look at https://travis-ci.org/voxpupuli/puppet-archive/jobs/124201892#L341 and fix the style errors?

thiagomarinho:
Done! :)

PR #156, opened 2016-05-05
Ruby 1.8.7 does not have options hash

edestecd:
Ruby 1.8.7 Net::HTTP.start does not accept a hash of options.
This change allows it to work in both 1.8.7 and newer rubies.

### DIFF
--- a/lib/puppet_x/bodeco/util.rb
+++ b/lib/puppet_x/bodeco/util.rb
@@ -41,7 +41,9 @@ module PuppetX
       end
 
       def follow_redirect(uri, option = { :limit => FOLLOW_LIMIT }, &block)
-        Net::HTTP.start(uri.host, uri.port, :use_ssl => (uri.scheme == 'https')) do |http|
+        req = Net::HTTP.new(uri.host, uri.port)
+        req.use_ssl = (uri.scheme == 'https')
+        req.start do |http|
           http.request(generate_request(uri)) do |response|
             case response
             when Net::HTTPSuccess


bastelfreak:
Hi,

we dropped ruby1.8.7 support. THis is a very very old and buggy and unsupported version. Please upgrade you ruby or use a older version of the module. If you need help for upgrading your puppet env please join our IRC channel #voxpupuli on freenode.

edestecd:
Puppet 3.x on RedHat 6 installed from puppetlabs rpms still uses Ruby 1.8.7.
If your module claims to support Pupept 3.x and RedHat 6, then it is advisable to support 1.8.7.

This is the only correction I have needed to make.  Why not support it, if possible?

bastelfreak:
We still support Ruby3.X, but not Ruby1.8.7. As I said this is a totally unpatched Version, it is slow, has open issues and is unmaintained. We had a huge discussion a while back about dropping it. If your OS vendor doesn't ship a decent Ruby version you can switch to puppetlabs AIO packages or build them on your own.

edestecd:
Puppetlabs AIO is puppet > 4, correct?   So there is no option for a newer ruby on Puppet 3.x, other than compiling it yourself.

edestecd:
What version of this module supports 1.8.7?

bastelfreak:
This is the case for Centos6, yes. RedHat is unable to ship a new version for $reasons :(
But Puppet 3 itself supports newer version, Centos7/Debian8 for example ship ruby2.0/2.1 which works well for Puppet. Archlinux runs Puppet3 on Ruby2.3.

---

I don't have in my mind which release still supports 1.8 :( I could be 0.4.8

nanliu:
I'm not opposed to this patch for backwards compatibility, besides the default provider is curl on linux.

rnelson0:
It's important to note that we can make changes that work with Ruby 1.8.7, but there will be no testing with 1.8.7 so there could be regressions in the future.

Re: EL6 and puppet 3, you can use the SCL repo or rvm/rbenv to get a newer version of Ruby for use with puppet.

@bastelfreak is the ruby_dep issue resolved in modulesync? If so, we should sync and then have this rebased.

edestecd:
@nanliu I had the same thought about curl being the default on linux.  I wondered why this code was being executed...  Does that code get used to fetch the checksum or something?

I have something like this:
```yaml
archives:
  /tmp/foo.txt:
    source: "https://example.edu/%{::environment}/Artifact/foo.txt"
    checksum_url: "https://example.edu/%{::environment}/Artifact/foo.txt.sha1"
    checksum_type: sha1
```

edestecd:
@rnelson0 yes Foreman uses the scl repos I believe to get ruby 1.9.3.  I may look into that...
But for now I need something to work on my boxes.  I understand that you do not test against 1.8.7 and support will be up to me until I can switch to a newer ruby...

bastelfreak:
New modulesync is happening in https://github.com/voxpupuli/puppet-archive/pull/157

bastelfreak:
@rnelson0 the gem issue is fixed now, I did a modulesync.

rnelson0:
@edestecd can you rebase this and push? CI should pass this time.

edestecd:
@rnelson0 the new commits convert everything to ruby 1.9 hash syntax, which makes my change moot, since I am trying to make 1.8.7 work.

rnelson0:
@edestecd I was afraid of that. The use of rubocop with a 1.9 compat level can be pretty strict about hash syntax and other changes, but not syncing this module with other voxpupuli modules is not a good position either. I'm sorry!

Is there any way forward that would permit compatibility with 1.8.7 without violating the modulesync settings? Would one or two rubocop disable comments resolve this or would it require dozens and defeat the point of rubocop?

edestecd:
This is the culprit:
https://github.com/voxpupuli/puppet-archive/commit/ccb0a26143a33139c25bba6c1c5336dd01a83f4d#diff-11a0d7906801d9dea0eccb85667ad811L45

rnelson0:
That ties in with
https://github.com/voxpupuli/puppet-archive/commit/ccb0a26143a33139c25bba6c1c5336dd01a83f4d#diff-11a0d7906801d9dea0eccb85667ad811R2.
How many 'violations' are we looking at now?

jyaworski:
@rnelson0 we would have to disable the hash_syntax cop again. I'm not against it, but I think we should leave it on for the spec tests. Maybe just leave it off for lib and templates.

bastelfreak:
I'm against disabling the cop. There are enough solutions to get a newer ruby version. 1.8.3 is EOL since a long long time and we should git rid of it as soon as possible.

telegrapher:
Hello guys!
Thanks for your work!! This is a great module!!

I've been bit by this issue. I've had to upgrade puppet-archive due other modules' dependencies and then all the CentOS 6 machines stopped working. This patch worked fine for me.

A Ruby upgrade will not be possible in many environments, a clear warning for RHEL/CentOS 6 users upgrading to >0.4.8 could be really useful!!

jyaworski:
This patch works, with no change in functionality. I'd be fine with accepting it. @edestecd can you rebase?

edestecd:
@jyaworski have you fixed rubocop to accept the old ruby 1.8.7 hash syntax?
Unless that is reverted 1.8.7 will not work and my pr is moot...

jyaworski:
@edestecd oh. You meant outside of the scope of this PR. No, we have not yet changed rubocop and we are unlikely to do so. I agree that 1.8.7 is still around (curse you RHEL6) and it's annoying to work around it, but we are discussing here in https://github.com/voxpupuli/plumbing/issues/36 about using softwarecollections on RHEL6 to get around this annoyance. Please add input there.

edestecd:
I'm up for learning how to make the puppetlabs rpms for puppet 3.x use the softwarecollections.
I have not figured it out yet...

Hopefully we can migrate to puppet 4 soon.  I'm waiting on full Foreman support, which should be fairly soon....

jyaworski:
@edestecd Foreman has very recently (last week-ish) supported puppet 4 in the nightlies. Next release.

I _think_ we can append scl ruby1.9.3 enable <command> to the puppet service file. @bastelfreak has some experience in this, but with Arch I think.

bastelfreak:
On arch it is a bit different, we currently ship ruby2.3 as the default MRI version, so we don't have to deal with SCL. Building puppet3 on ruby2.1 or older worked out of the box, for ruby2.2 and 2.3 we need  a oneline patch.

bastelfreak:
@edestecd how should we proceed here? Are you going to upgrade your puppet/ruby version and this can be closed or did I understand you wrong?

edestecd:
@bastelfreak I think we will upgrade to puppet 4 when the next version of Foreman comes out.
Until then we can use my forked version...

PR #157, opened 2016-05-06
modulesync 0.5.1

bastelfreak:


igalic:
the jruby failures are unrelated to this patch

PR #158, opened 2016-05-06
Skip jruby for provider tests.

nanliu:


jyaworski:
Jruby not supported? huh?

nanliu:
JRuby on the master is ok, afaik there no jruby puppet agent so we can ignore these failed tests for the provider.

jyaworski:
Bizarre. OK. Let's hope the puppet agents don't switch to jruby.

PR #164, opened 2016-05-12
Update documentation

jhg03a:
Update documentation for handling situations involving a ```puppet://``` url and better local file usage documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@
 3. [Setup](#setup)
 4. [Usage](#usage)
    * [Example](#usage-example)
+   * [Puppet URL](#puppet-url)
    * [File permission](#file-permission)
    * [Network files](#network-files)
    * [Extract customization](#extract-customization)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,6 +132,29 @@ archive { '/tmp/test100k.db':
 }
 ```
 
+### Puppet URL
+
+Currently there isn't support for using ```puppet:\\``` style url in the source parameter.  It is still possible, but it would be done in two steps:
```
nanliu: Wrong slashes puppet:///, and please note do not enable  `cleanup=>true` which is not idempotent with this work around.

README.md
```diff
@@ -131,6 +132,29 @@ archive { '/tmp/test100k.db':
...
+  ensure => file,
+  source => "puppet:///modules/profile/${docs_filename}",
+}
+
+archive { $docs_gz_path:
+  path          => $docs_gz_path,
+  extract       => true,
+  extract_path  => $homedir,
+  creates       => "${homedir}/help" #directory inside tgz
+  require       => [ File[$homedir], File[$docs_gz_path] ],
```
nanliu: The $homedir is not necessary, it should be picked up by autorequire:

https://github.com/voxpupuli/puppet-archive/blob/master/lib/puppet/type/archive.rb#L164-L165
```
      self[:extract_path],
```

README.md
rnelson0: Hate to sound like a pedant, but the first part of the sentence should be "Currently there is no support for using a .." - important for ESL readers.

README.md
```diff
@@ -131,6 +132,29 @@ archive { '/tmp/test100k.db':
...
 ```
 
+### Puppet URL
+
+Currently there isn't support for using ```puppet:\\``` style url in the source parameter.  It is still possible, but it would be done in two steps:
+
+```puppet
+$docs_filename = 'help.tar.gz'
+$docs_gz_path  = "/tmp/${docs_filename}"
+$homedir = '/home/myuser/'
```
rnelson0: Before getting into the specifics of usage, describe what the goal is, i.e. to place the file `help.tar.gz` from the `profile` module's `files` directory at `/tmp/help.tar.gz`, and then expand that file in the user's home directory, `/home/myuser`. I assume that is correct, anyway.

bastelfreak:
@jhg03a thanks for the PR! Can you please take a look at the comments above?

Issue #166, opened 2016-05-18
change in puppet 4.5.0 causing archive type to fail when extract_path not used


hdeadman:
I think something may have changed in the newly release Puppet 4.5.0 that makes this statement from 
archive.rb fail:

```
  autorequire(:file) do
    [
      Pathname.new(self[:path]).parent.to_s,
      self[:extract_path],
      '/root/.aws/config',
      '/root/.aws/credentials',
    ]
  end
```
It fails if the extract_path parameter is not used (which is the case in the rtyler/jenkins plugin.pp file. When it fails the error message is as follows:
```
 Error: Failed to apply catalog: No title provided and :file is not a valid resource reference
```

There is no indication of where in the catalog that error is coming from so it is painful to track down. 

The fix is to make the       autorequire of a file named self[:extract_path] conditional on that parameter being defined.

petems:
@hdeadman if you run puppet with `--trace` does it show you where the error is coming from?

hdeadman:
Here is the trace but it doesn't help too much. It does show the problem coming from an autorequire. I know the cause is the self[:extract_path] line because I can comment that out and it will run. I am invoking archive indirectly through [jenkins::plugin{'git': }](https://github.com/jenkinsci/puppet-jenkins/blob/master/manifests/plugin.pp) and the error only happens if the jenkins plugin is not already installed. 

```
==> services1: Notice: Compiled catalog for services1.redacted in environment local in 39.74 seconds
==> services1: Error: Failed to apply catalog: No title provided and :file is not a valid resource reference
==> services1: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/resource.rb:544:in `extract_type_and_title'
==> services1: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/resource.rb:529:in `type_and_title'
==> services1: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/resource/catalog.rb:353:in `resource'
==> services1: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type.rb:2138:in `block (2 levels) in autorelation'
==> services1: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type.rb:2134:in `each'
==> services1: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type.rb:2134:in `block in autorelation'
[...]

PR #167, opened 2016-05-19
avoid creating file autorequire with nil variable

hdeadman:
This is a fix for #166.

### DIFF
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -165,7 +165,7 @@ Puppet::Type.newtype(:archive) do
       self[:extract_path],
       '/root/.aws/config',
       '/root/.aws/credentials',
-    ]
+    ].compact
   end
 
   autorequire(:exec) do


lib/puppet/type/archive.rb
```diff
@@ -160,12 +160,16 @@ def should_to_s(value)
...
     [
       Pathname.new(self[:path]).parent.to_s,
       self[:extract_path],
       '/root/.aws/config',
       '/root/.aws/credentials',
-    ]
+    ].each do |file| 
+      autos << file unless file.nil?
+    end
+    autos
```
nanliu: Please just do .compact on the array. Also we might consider only adding /root/.aws on posix platforms.

nanliu:
Would appreciate a spec test supporting this change.

hdeadman:
I updated to use compact. I don't have much Ruby experience but I figured there was a simple way to do it. I am not currently setup to run spec tests, I have only ever written one, but would adding Puppet 4.5 to the travis text matrix cause many of the existing tests to fail (since they don't set the extract_path argument)? The error happens during catalog execution (vs compilation), do tests involve executing the catalog?

rnelson0:
Unit tests (rspec-puppet) are catalog only, testing the state immediately
after compilation. Acceptance tests (beaker) can be used to test the
results of execution.

-- 

Rob Nelson
rnelson0@gmail.com

jyaworski:
Hi @hdeadman. As @rnelson0 mentioned our acceptance testing is a bit sub-par right now. We're actively looking into how to improve this.

That being said, I can accept this PR without a spec test due to its nature. Can you squash your commits?

PR #168, opened 2016-05-26
Update from voxpupuli modulesync_config

jyaworski:


jyaworski:
Merging so tests can pass again...

PR #176, opened 2016-06-21
make our type camptocamp/archive compatible

igalic:
our type now accepts the same parameters as camptocamp/archive and processes them correctly.

### DIFF
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -139,9 +139,11 @@ Puppet::Type.newtype(:archive) do
 
   newparam(:checksum) do
     desc 'archive file checksum (match checksum_type).'
-    newvalues(%r{\b[0-9a-f]{5,128}\b}, :true, :false)
+    newvalues(%r{\b[0-9a-f]{5,128}\b}, :true, :false, :undef, nil, '')
     munge do |val|
-      if val == :true || val == :false
+      if val.nil? || val.empty? || val == :undef
+        :false
+      elsif val == :true || val == :false
         resource[:checksum_verify] = val
       else
         val
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -76,6 +76,17 @@ Puppet::Type.newtype(:archive) do
       end
     end
   end
+  newparam(:target) do
+    desc 'target folder path to extract archive. (this parameter is for camptocamp/archive compatibility)'
+    validate do |value|
+      unless Puppet::Util.absolute_path? value
+        raise ArgumentError, "archive extract_path must be absolute: #{value}"
+      end
+    end
+    munge do |val|
+      resource[:extract_path] = val
+    end
+  end
 
   newparam(:extract_command) do
     desc "custom extraction command ('tar xvf example.tar.gz'), also support sprintf format ('tar xvf %s') which will be processed with the filename: sprintf('tar xvf %s', filename)"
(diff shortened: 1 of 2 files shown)


lib/puppet/type/archive.rb
```diff
@@ -109,24 +120,64 @@ def should_to_s(value)
...
+    munge do |val|
+      resource[:checksum_verify] = val if val == :true || val == :false
+    end
+  end
+  newparam(:digest_string) do
+    desc 'archive file checksum (match checksum_type)
+    (this parameter is for camptocamp/archive compatibility).'
+    newvalues(/\b[0-9a-f]{5,128}\b/)
+    munge do |val|
+      resource[:checksum] = val
```
igalic: this munge needs to be handled differently

daenney:
So what's the goal of achieving compatibility?

igalic:
currently, it's impossible to drop-in-replace either of the modules with the other
This patch serves as an attempt to get them a little bit closer.

lesaux:
I've recently deprecated installing kibana4 from the tar.gz archive, since elastic.co is now maintaining packages.

igalic:
well that's one problem solved… 1209012390439034 others to go ;)

igalic:
this commit now adds `archive::download`

alexjfisher:
I've been wondering whether we really want to clutter this module with compatibility types and parameters.

If we get the missing features implemented, might it be better to then start creating PRs against popular modules that use camptocamp/archive? It's even possible to make a module work with *both* camptocamp/archive and puppet/archive by using stdlib's `load_module_metadata`.  That's what I've done in https://github.com/cpitman/puppet-database_schema/pull/4

igalic:
[from Joe Duffy's blog about Midori](http://joeduffyblog.com/2016/02/07/the-error-model/#advanced-type-systems) (a research operating system)

>  Lets say that for any given 1 producer of state there are 10 consumers. Rather than having each of those 10 defend themselves against error conditions, we can push the responsibility back onto that 1 producer, and either require a single assertion that coerces the type, or even better, that the value is stored into the right type in the first place.

@alexjfisher what you're suggesting is "worse is better", i disagree.

jyaworski:
I agree with @igalic. We want to fix the module, not the things that use it. Take the Linux stance: 'userspace is never the problem.' I'm for adding compatibility types. We can remove them in a major release later.

alexjfisher:
... turns out, I can ;)
These are the forge modules that list camptocamp/archive as a dependency in their latest release.
Ordered by total downloads across all releases.
```
[{:module=>"rtyler-jenkins", :downloads=>253718},
 {:module=>"bfraser-grafana", :downloads=>47516},
 {:module=>"camptocamp-tomcat", :downloads=>26947},
 {:module=>"tohuwabohu-duplicity", :downloads=>8957},
 {:module=>"tohuwabohu-drupal", :downloads=>6795},
 {:module=>"pennycoders-marathon", :downloads=>6392},
 {:module=>"jschilperoord-papertrail", :downloads=>6324},
 {:module=>"pennycoders-zookeeper", :downloads=>6078},
 {:module=>"pennycoders-mesos", :downloads=>5148},
 {:module=>"nrvale0-google_cloud_sdk", :downloads=>4108},
 {:module=>"cpitman-database_schema", :downloads=>4020},
 {:module=>"tohuwabohu-roundcube", :downloads=>3956},
 {:module=>"tohuwabohu-confluence", :downloads=>3790},
 {:module=>"flypenguin-teamcity", :downloads=>3137},
 {:module=>"benforge-activemq", :downloads=>2021},
 {:module=>"benforge-jbosseap", :downloads=>1740},
 {:module=>"GobuNatarajan-gsutil", :downloads=>1510},
 {:module=>"puppetlabs-rkt", :downloads=>1423},
 {:module=>"RanjithKumar45-gcloudsdk", :downloads=>0}]
```

igalic:
nice! thank you for the stats!

igalic:
okay, so, in my `archive::download` patch i made some adjustments to fix up incompatibilities… which makes me wonder what i can do to `archive` itself to discover usage of camptocamp style sematics vs voxpupuli style.

i'm also contemplating to copy camptocamp's spec tests to see how compatible we are — but not sure how that'll actually work, since they are written for a defined type, and ours is a type/provider

igalic:
someone tried using this patchset as replacement for camptocamp/archive and reported the following error:
```
Error: Failed to apply catalog: Parameter checksum failed on Archive[/var/lib/jenkins/plugins/workflow-scm-step.hpi]: Invalid value "". Valid values are true, false. Valid values match /\b[0-9a-f]{5,128}\b/. at /etc/puppetlabs/code/environments/dev/modules/archive/manifests/download.pp:54
```

igalic:
tests green except for that blacksmith/rest-client issue, for which we have a pr lined up here https://github.com/voxpupuli/puppet-blacksmith/pull/35

lib/puppet/type/archive.rb
```diff
@@ -109,24 +120,74 @@ def should_to_s(value)
...
+    end
+  end
+
   newparam(:cookie) do
     desc 'archive file download cookie.'
   end
 
   newparam(:checksum) do
     desc 'archive file checksum (match checksum_type).'
+    newvalues(%r{\b[0-9a-f]{5,128}\b}, :true, :false, :undef, nil, '')
```
igalic: this is awful

igalic: but it works with the latest ***release*** or rtyler/jenkins

igalic: (a.k.a.: we're running it in prod now ;)

raphink:
Camptocamp compatible

manifests/download.pp
```diff
@@ -0,0 +1,64 @@
+# == Definition: archive::download
```
alexjfisher: @igalic Maybe document this type as for compatibility with camptocamp?  You would otherwise always use the native `archive` type directly?

alexjfisher:
Other than a updating the README and comments in archive::download LGTM.  :+1:

manifests/download.pp
igalic: it wasn't documented in camptocamp/archive either ~_~
i don't *want* to document it, because it's just a shim. private api that moves as camptocamp/archive moves.

alexjfisher: sorry, I wasn't clear (sorry of my life!) I don't mean document all the parameters etc.  I mean just add a line to say that it is a shim/private etc.  Something to discourage new users from thinking this is the way they should be using the module.

alexjfisher: similar to how the new parameters in the type are marked `(this parameter is for camptocamp/archive compatibility)`

Issue #175, opened 2016-06-21
Inconsistent cacerts used


alexjfisher:
As observed on puppet 4.5 AIO on OEL7....

On posix, the default provider is `curl`.  When the type invokes curl to download a file from an HTTPS server, curl uses the system cacerts (on OEL7 at least).

But when the checksum file is downloaded, the curl provider doesn't implement this download, instead relying on it's parent provider methods.  The `ruby` provider doesn't use the system ca certs and this can lead to failures to download the checksum file even when the main download completed successfully.  End result when trying to download large files is a very full filesystem.  Each 30 minutes a new temporary file is downloaded to /tmp and is never cleaned up...

igalic:
i suggest three fixes for this

* ensure that the checksum is downloaded through the same means as the file itself
* cleanup temporary files after failure. Failure should be LOUD
* document the location of CA files that need updating when updating the system CAs (for AIO agent that would be `/opt/puppetlabs/puppet/ssl/puppet-cacerts` & /opt/puppetlabs/puppet/ssl/cert.pem`)

alexjfisher:
Implementing https://github.com/voxpupuli/puppet-archive/issues/174 would help too.

alexjfisher:
Fixed in https://github.com/voxpupuli/puppet-archive/pull/179 Thanks @nanliu !

PR #179, opened 2016-06-26
GH-175 Fix provider checksum content fetch method.

nanliu:
If a user switch providers, we should use the same provider to obtain
checksum content.


igalic:
nice!

PR #181, opened 2016-07-12
Update metadata.json dependencies and requirements

alexjfisher:
`Puppet::Parameter::Boolean` isn't available before `puppet 3.3`.
`delete_undef_values()` function was introduced in `stdlib 4.2.0`.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -33,8 +33,7 @@
         "4",
         "5",
         "6",
-        "7",
-        "8"
+        "7"
       ]
     },
     {
@@ -91,7 +90,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">=3.0.0 < 5.0.0"
+      "version_requirement": ">= 3.4.0 < 5.0.0"
     }
   ],
   "name": "puppet-archive",
@@ -106,7 +105,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 3.0.0 < 5.0.0"
+      "version_requirement": ">= 4.2.0 < 5.0.0"
     }
   ]
 }


igalic:
@alexjfisher i think a bump of stdlib is okay, but we should stay compatible with older versions of puppet for *this* 1.0.0 release

alexjfisher:
This PR was to make sure that the metadata matched reality.

It was my https://github.com/voxpupuli/puppet-archive/pull/180 PR that broke puppet < 3.3.  We only run tests against latest puppet 3 and latest puppet 4, so this makes it difficult to check that earlier versions still work.  It is possible to test locally, but you have to hack the Gemfile first.  eg `puppet-strings` requires puppet 3.7.

Puppet 3.3 was released September 12, 2013.  That said, I could rework https://github.com/voxpupuli/puppet-archive/pull/180 if required.

It's https://github.com/voxpupuli/puppet-archive/blob/master/manifests/nexus.pp that requires stdlib 4.2.  I guess somebody could change the code, but requiring a fairly recent stdlib isn't as onerous as requiring newer versions of puppet.

alexjfisher:
The type doesn't work with puppet <3.3 because I used https://docs.puppet.com/guides/custom_types.html#boolean-parameters

alexjfisher:
1.0.0 is already dropping support for ruby 1.8.  Personally, I think this is going to cause more users problems than requiring a very slightly later puppet version.

bastelfreak:
Could we drop everything puppet <3.8? These versions are packaged by puppet so everybody should easily use them (if he doesn't already). Also we only test on this version and not on older ones. If we say we support for example 3.4, we should test every minor release after that?

alexjfisher:
I think we should at least try to test each module's minimum supported puppet version.  Then I wouldn't have accidentally broken puppet 3.0-3.2 in my last PR.  I think puppet 3.x was relatively stable from about 3.4 onwards.

nibalizer:
Maintaining back to 3.4 is what I'd like to see. That was the version that shipped with trusty and will be around forever.

nibalizer:
This looks good to me as written.

bastelfreak:
but do we want it to be around forever? we're only encouraging people to not update when we support it.

igalic:
no, at one point or another, we can make breaking changes and call that 2.0.x — but i don't wanna do that *now*

bastelfreak:
oookay

bastelfreak:
so we've to bump it in this PR to 3.4?

igalic:
i'm not entirely sure i like this idea, since archive has been around a long time, and is used a lot, it's now Camptocamp compatible, and so we shouldn't screw over people who have to use it with older versions of puppet just because of booleans.

this is a new and strange and conservative stance for me, and i'm fine with y'all disagreeing with me. it's especially strange since i have been working in 4-only environments for about half a year :D

bastelfreak:
still nobody is forced to upgrade their existing modules. Old releases work fine in old environments.

igalic:
true, but people will adapt their version requirements!

igalic:
it'lll be all like: LOOK IT'S 1.0.x! STABLE! UPGRADE!

bastelfreak:
:hankey:

alexjfisher:
@bastelfreak Suppose if 3.4 is the lowest version we definitely want to keep it working with going forward.  I still wouldn't bump the metadata if we know it works with 3.3.  If someone later made a change that broke 3.4, that'd be allowed and we'd bump the metadata to match then?

@igalic Yeah, booleans *can* be removed to keep the module working with even earlier versions.  In general, I wouldn't want to make too many compromises though.  After all, how far back would we want to go?  I don't actually know for certain it did work with puppet 3.0 - 3.2, all I know is it definitely doesn't anymore.

igalic:
release with 3.4 as is: :+1:

alexjfisher:
@igalic OK.  I've pushed a new commit with 3.4.

PR #186, opened 2016-07-15
Move cacert to pluginsync directory.

nanliu:


rigareau:
The cacert.pem file is very old.  Why not update it with the new version that is availlable at https://curl.haxx.se/ca/cacert.pem

igalic:
:+1:

Issue #191, opened 2016-08-03
extract_flag doesn't work


tibers:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8.7
- Ruby: ruby 1.8.7 (2013-06-27 patchlevel 374) [x86_64-linux]
- Distribution: Red Hat Enterprise Linux Server release 6.7 (Santiago)
- Module version: 1.0.0

### How to reproduce (e.g Puppet code you use)

```puppet
archive { "/tmp/reporting-${reporting_tag}-report-generation.zip":
  ensure       => present,
  cleanup      => true,
  creates      => "${deploy_reporting_base_dir}/filter_FE_R_REPORT_${reporting_tag}.mvel",
  source       => "${deploy_reporting_base_url}/${reporting_tag}/reporting-${reporting_tag}-report-generation.zip",
  extract_path => $deploy_reporting_base_dir,
  extract_flag => '-of',
  extract      => true,
}
```

### What are you seeing

Error: Could not retrieve catalog from remote server: Error 400 on SERVER: no parameter named 'extract_flag' at /etc/puppetlabs/code/environments/preprod/modules/anonymousreportingthing/manifests/deploy_reporting_tag.pp:9 on Archive[/tmp/reporting-1.1.28-report-generation.zip] at /etc/puppetlabs/code/environments/preprod/modules/anonymousreportingthing/manifests/deploy_reporting_tag.pp:9 on node somenodehere.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run

### What behaviour did you expect instead

Do the work.

### Any additional information you'd like to impart

Followed documentation here: https://forge.puppet.com/puppet/archive#extract-customization

nanliu:
Ouch, bad README documentation. The option should be `extract_flags` as seen here:

https://github.com/voxpupuli/puppet-archive/blob/master/manifests/nexus.pp#L80

tibers:
Alright my bad for not reading the code too - but it's a quick fix: 
https://github.com/voxpupuli/puppet-archive/pull/192

nanliu:
Fixed by #192.

PR #192, opened 2016-08-03
Issue #191 - fixing README

tibers:
And it's my bad for not reading the code but I was so excited I couldn't help but cut and paste. ;)


bastelfreak:
Thanks for the patch! can you please rebase after https://github.com/voxpupuli/puppet-archive/pull/193 is merged? Then travis should be green here again.

Issue #194, opened 2016-08-04
AIX - extraction fails


winningpanthers:
AIX does not have the GNU tar, tar -xzvf does not work in AIX and extraction fails. Can you please fix the issue?

### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet:
- Ruby:
- Distribution:
- Module version:

bastelfreak:
Hi @winningpanthers, thanks for reporting this. are you able to provide a patch or a PR for this?

winningpanthers:
You're welcome. Still working on it. Will let you know.

nanliu:
I'm guessing this should work like Solaris. Let me know if the fix works on AIX or if we should use a different set of commands since I don't have any AIX hardware to verify.

winningpanthers:
Yes. thank you. I fixed this on my local and was about to upload the details. This works. Tested this on my AIX. Appreciate your help.

PR #197, opened 2016-08-09
Fix GH-194 AIX extraction

nanliu:
* AIX extraction should work similar to Solaris
* Add unit test coverage.


igalic:
i am surprised we don't need any special casing for freebsd.

PR #202, opened 2016-08-18
Syntax fix for ruby 1.8

dconry:
I know we do not support ruby 1.8 anymore, but this seems like a very small fix; a similar fix was made for puppet-staging (see second change in link below):

https://github.com/voxpupuli/puppet-staging/commit/6729b2a313e647af881d26f0e11b4f7fb7432b59

I've patched this locally, as I still have a great many CentOS 6 machines to support.  Would be nice to get it into upstream for next release, if no objections.

### DIFF
--- a/lib/facter/archive_windir.rb
+++ b/lib/facter/archive_windir.rb
@@ -1,5 +1,5 @@
 Facter.add(:archive_windir) do
-  confine osfamily: :windows
+  confine :osfamily => :windows
   setcode do
     program_data = `echo %SYSTEMDRIVE%\\ProgramData`.chomp
     if File.directory? program_data


bastelfreak:
In staging we backported this. I think we have to merge this into a 0.X branch which is a bit painful. Can't you use an older version of the module or upgrade to puppet4 or to a newer CentOS? We're happy to assist you in all of the three steps, you can reach us in #voxpupuli on freenode.

dconry:
Ran into this because rtyler/jenkins changed deps to use this instead of the camptocamp version.

I believe this change won't break any upstream puppets/rubies, but if I'm wrong or it's otherwise too difficult to merge, I can just maintain my local patch (upgrading CentOS and/or to puppet4 everywhere is more than I can handle, at the moment).  In that case, feel free to close to request.

Thanks!

bastelfreak:
well it is doable, the point here is that ruby187 is EOL since a very very long time. There are open vulns that won't be fixed anymore upstream. People should really upgrade to something recent. From my personal point of view I can't support something that is in such a broken state.

---

I will leave this open for 72hrs. Maybe one of the other contributors have a different opinion.

dconry:
Understood, thanks.  Only a minor inconvenience on my side to maintain the patch locally.

RH is supposed to backport security fixes, but I run into "old OS & libraries vs new software" a great deal... not fun.  Might have to separate my CentOS 6 instances into a different environment.

nanliu:
Are you sure that's the only change required? There are other places such as: `.provide(:curl, parent: :ruby)`

dconry:
I'm sure this is the only change needed to get the Ruby library to not return parse errors on CentOS 6/Puppet 3.8.7/Ruby 1.87.  I've not tried actually using the code on such a system, but for me it's sufficient that the distributed library doesn't cause errors during manifest compile on such systems.

nanliu:
Ah, so that gets around loading the facts, but the module is still not usable. Not sure, I would either aim for full 1.8.7 compatibility or leave it as is instead of a partial patch.

alexjfisher:
Making the fact not blow up on ruby 1.8 can sometimes be helpful by itself.  For example, if in your environment you have both Windows (ruby > 1.8) and Centos 6 (ruby 1.8) boxes but you are only wanting to use puppet-archive in one of your windows role.  Also see https://github.com/voxpupuli/puppet-iis/pull/111

But I also think we should be open to a final 0.6 'sunset' release of this module with ruby 1.8 support for those who still need it.

dconry:
I can confirm that actually applying the module with a ruby 1.8.7 based agent would require further (apparently non-trivial) fixes.  I can look into discovering these & adding them to the patch, if it would be accepted.

For us it is sufficient to have to fact load without error on such systems...  this module is only in actual use on systems with much newer OS versions.

rnelson0:
I worry about a patch that promises 1.8.7 support when we do no testing of it. We cannot guarantee even today that it will work, and we will have no idea if regressions occur. I am against this PR on master but would be open to a PR against a 0.x tag if it's required to support one last pre-1.0 version for some reason. Is that really necessary, though?

jyaworski:
I don't support this. We don't offer 1.8.7 support, and anything that works on it is by accident rather than by design. Please pin modules to compatible versions.

alexjfisher:
I propose making the fact 1.8 compatible on master.  This is trivial and useful for people with environments containing old and new rubies such as @dconry.

I'd also be happy to do the work to create a 0.x branch where a final pre 1.0 release could be made.  Having one camptocamp/archive compatible release that works with ruby 1.8 would be useful to quite a few people (including myself actually).

rnelson0:
I'm comfortable with that work *within* the 0.x branch, not in 1.x (master).

alexjfisher:
The trouble I've got with using an older version is that my recently rebuilt puppet masters (on OEL7) use puppet-archive too.  I've got other roles that also use it and need the camptocamp compatible version. (thanks to @igalic BTW!)
The older (ruby 1.8) systems I have to support with puppet 3 can't be pinned to a ruby 1.8 compatible version as mismatching versions between masters and agents [doesn't work.](https://tickets.puppetlabs.com/browse/SERVER-94)

alexjfisher:
@dconry https://github.com/voxpupuli/puppet-archive/pull/215 was merged.  It *doesn't* make the type usable on your Centos 6 systems, but it will stop the `archive_windir` fact from generating errors on those systems.

Issue #204, opened 2016-08-29
archive::nexus ignore changes


myMarck:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8.5
- Ruby: 1.9.3p484
- Distribution: Ubuntu 14.0.4
- Module version: 1.1.1

### How to reproduce (e.g Puppet code you use)
Run the example code in nexus::archive

   archive::nexus { '/tmp/jtstand-ui-0.98.jar':
     url        => 'https://oss.sonatype.org',
     gav        => 'org.codehaus.jtstand:jtstand-ui:0.98',
     repository => 'codehaus-releases',
     packaging  => 'jar',
     extract    => false,
   }

ls -al /tmp/jtstand-ui-0.98.jar 
-rw-r----- 1 root root 286643 Aug 29 12:02 /tmp/jtstand-ui-0.98.jar

change the file size to zero by doing

sudo rm /tmp/jtstand-ui-0.98.jar && sudo touch /tmp/jtstand-ui-0.98.jar

run puppet again

### What are you seeing

ls -al /tmp/jtstand-ui-0.98.jar 
-rw-r--r-- 1 root root 0 Aug 29 12:04 /tmp/jtstand-ui-0.98.jar

### What behaviour did you expect instead

I would expect the archive module to replace the file.

### Output log
NA

### Any additional information you'd like to impart
It works in version 0.5.1
This problem also affects SNAPSHOT dependencies that are not replaced when new archives exsists in the repository

nanliu:
@myMarck, check and see if #211 resolves this issue

PR #211, opened 2016-08-31
Fix GH-204 resolve camptocamp archive regression

nanliu:
The camptocamp archive digest_type default of :none overwrote puppet
archive checksum_type value, so even when we have a valid checksum such
as md5, the digest_type overwrote to :none. This results in files with
invalid checksums not being replaced. This PR addresses this bug.

### DIFF
--- a/lib/puppet/provider/archive/curl.rb
+++ b/lib/puppet/provider/archive/curl.rb
@@ -28,17 +28,15 @@ Puppet::Type.type(:archive).provide(:curl, parent: :ruby) do
   end
 
   def remote_checksum
-    @remote_checksum ||= begin
-      params = curl_params(
-        [
-          resource[:checksum_url],
-          '-fsSL',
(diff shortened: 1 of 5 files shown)


nanliu:
Verified by testing end user scenario of rm && touch of the file in question:

```
Notice: Compiled catalog for centos72.shared in environment production in 0.21 seconds
Notice: /Stage[main]/Main/Archive::Nexus[/tmp/jtstand-ui-0.98.jar]/Archive[/tmp/jtstand-ui-0.98.jar]/ensure: replace archive: /tmp/jtstand-ui-0.98.jar from (md5)d41d8cd98f00b204e9800998ecf8427e to (md5)68316c3605e197c3c3eb1630c441d997
```

alexjfisher:
LGTM.  I've tested this PR with all 3 providers and in fixes the issue (which I can replicate) for all of them.

Issue #213, opened 2016-08-31
`allow_insecure` not working in Windows


alexjfisher:
I'm not sure if it's specifically Windows, (where I run puppet 3.8 with puppetlabs shipped ruby 2.0.0), or the ruby provider in general, but https://github.com/voxpupuli/puppet-archive/pull/198 doesn't seem to work for me.

What does work is if I move the `verify_mode = OpenSSL::SSL::VERIFY_NONE` into the 3rd parameter of `Net::HTTP.start` (the hash where only `use_ssl` is currently set)

alexjfisher:
Also not working in Linux when using the `ruby` provider.
```
alex@alex-VirtualBox:~$ puppet --version
4.6.1
alex@alex-VirtualBox:~$ facter -p rubyversion
2.3.1
alex@alex-VirtualBox:~$ uname -a
Linux alex-VirtualBox 4.4.0-28-generic #47-Ubuntu SMP Fri Jun 24 10:09:13 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
alex@alex-VirtualBox:~$ cat test.pp
archive {'/tmp/master.zip':
  source         => 'https://192.30.253.113/voxpupuli/puppet-archive/archive/master.zip',
  provider       => 'ruby',
  allow_insecure => true,
}
alex@alex-VirtualBox:~$ puppet apply -t test.pp
Info: Loading facts
Info: Loading facts
Notice: Compiled catalog for alex-virtualbox.ordsvy.gov.uk in environment production in 0.14 seconds
Info: Applying configuration version '1472635763'
Error: Could not set 'present' on ensure: hostname "192.30.253.113" does not match the server certificate at 1:/home/alex/test.pp
Error: Could not set 'present' on ensure: hostname "192.30.253.113" does not match the server certificate at 1:/home/alex/test.pp
Wrapped exception:
hostname "192.30.253.113" does not match the server certificate
Error: /Stage[main]/Main/Archive[/tmp/master.zip]/ensure: change from absent to present failed: Could not set 'present' on ensure: hostname "192.30.253.113" does not match the server certificate at 1:/home/alex/test.pp
Notice: Applied catalog in 0.68 seconds
```

PR #214, opened 2016-08-31
Fix `allow_insecure` for ruby provider

alexjfisher:
Fixes https://github.com/voxpupuli/puppet-archive/issues/213

### DIFF
--- a/lib/puppet_x/bodeco/util.rb
+++ b/lib/puppet_x/bodeco/util.rb
@@ -24,7 +24,7 @@ module PuppetX
         @username = options[:username]
         @password = options[:password]
         @cookie = options[:cookie]
-        @insecure = options[:allow_insecure]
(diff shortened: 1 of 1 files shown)


lib/puppet_x/bodeco/util.rb
```diff
@@ -42,8 +42,13 @@ def generate_request(uri)
       end
 
       def follow_redirect(uri, option = { limit: FOLLOW_LIMIT }, &block)
-        Net::HTTP.start(uri.host, uri.port, use_ssl: (uri.scheme == 'https')) do |http|
-          http.verify_mode = OpenSSL::SSL::VERIFY_NONE if @insecure
```
alexjfisher: `Net::HTTP.start` fails due to certificate verify error and the code in the block (where we attempt to ignore cert errors) is never executed.

Issue #221, opened 2016-09-24
Some Gem dependencies require Ruby > 2.1


alexharv074:
When running a bundle install I'm getting Gem dependency issues.

### Affected versions/distributions

- Ruby: 2.0.0p481

### How to reproduce

```
$ bundle install
```

### Actual behaviour

```
Installing ruby_dep 1.4.0

Gem::InstallError: ruby_dep requires Ruby version >= 2.2.5, ~> 2.2.
```

and then later

```
Installing listen 3.1.5

Gem::InstallError: listen requires Ruby version >= 2.2.3, ~> 2.2.
```

### Expected behaviour

All Gems should install.

bastelfreak:
Hi,
We are aware of the issue. Ruby2.1 is kind-of EOL and only gets security updates.However you should still be able to install needed tests with this command:
```
bundle install --without system_tests development
```

alexharv074:
I could be out of date, but isn't Puppet itself shipped with Ruby 2.1.x?

bastelfreak:
The puppet-aio packages ship a bundled ruby21, yeah. But the required gems for testing have nothing to do with a running puppet environment.

bastelfreak:
Why do you want to install the gems? Do you want to run the tests? We created the gem groups with much care to ensure that they also work on ruby193 and everything newer.

alexharv074:
Yes, to run the tests.  As a Puppet developer, I install Ruby 2.1 everywhere for consistency with Puppet.  I would think most people do this?

bastelfreak:
Personally I use system-ruby which is currently ruby231. Our spec tests are designed to work with the provided bundler snippet I posted above. Only the rubocop test/raketask requires ruby231.

alexjfisher:
@bastelfreak I've just confirmed that all the gems needed to run the spec tests still install fine on ruby 1.9 (when using `bundle install --without system_tests development`)

@alexharv074 The module is tested against several different versions of ruby. Take a look in the https://github.com/voxpupuli/puppet-archive/blob/master/.travis.yml to see what we currently test against.  Are you using rvm?

alexharv074:
It's no big deal.  I already worked around it.  I just thought most Puppet programmers would set up their environments to please Puppet.  Feel free to decline the PR.

PR #222, opened 2016-09-24
Pin gem versions of ruby_dep and listen

alexharv074:


bastelfreak:
as discussed in https://github.com/voxpupuli/puppet-archive/issues/221#issuecomment-249368848

PR #223, opened 2016-09-25
[WIP] adding ensure => latest for s3

alexharv074:
Hi all, not sure if anyone has time but I'm struggling a little to add the feature I need, namely `ensure => latest` on s3 downloads (with a view to this being extended later).  Before I go too far, am I on the right track here?

### DIFF
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -31,7 +31,11 @@ Puppet::Type.type(:archive).provide(:ruby) do
   end
 
   def create
-    transfer_download(archive_filepath) unless checksum?
+    if resource[:ensure] == 'latest' and resource[:source] =~ /^s3/
+      transfer_download(archive_filepath) unless s3_checksum?
+    else
+      transfer_download(archive_filepath) unless checksum?
+    end
     extract
     cleanup
   end
@@ -92,6 +96,12 @@ Puppet::Type.type(:archive).provide(:ruby) do
     else
       archive_exist
     end
+  end
+
+  # Private: See if remote archive checksum matches.
+  # returns boolean
+  def s3_checksum?(remote_path)
+    # TODO
   end
 
   def cleanup
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -7,9 +7,9 @@ Puppet::Type.newtype(:archive) do
   @doc = 'Manage archive file download, extraction, and cleanup.'
 
   ensurable do
-    desc 'whether archive file should be present/absent (default: present)'
+    desc 'whether archive file should be present/latest/absent (default: present)'
 
-    newvalue(:present) do
+    newvalues(:present, :latest) do
       provider.create
     end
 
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/archive/ruby.rb
```diff
@@ -31,7 +31,11 @@ def exists?
   end
 
   def create
-    transfer_download(archive_filepath) unless checksum?
+    if resource[:ensure] == 'latest' and resource[:source] =~ /^s3/
+      transfer_download(archive_filepath) unless s3_checksum?
+    else
+      transfer_download(archive_filepath) unless checksum?
+    end
```
nanliu: Don't change it here, either modify the default checksum? method or implement your own s3 provider with your custom checksum? method.

lib/puppet/type/archive.rb
```diff
@@ -7,9 +7,9 @@
   @doc = 'Manage archive file download, extraction, and cleanup.'
 
   ensurable do
-    desc 'whether archive file should be present/absent (default: present)'
+    desc 'whether archive file should be present/latest/absent (default: present)'
 
-    newvalue(:present) do
+    newvalues(:present, :latest) do
```
nanliu: Do not add latest to ensurable method. This module logic for updating the files is simply:

1. is ensure == present
2. does checksum parameter match expectation?

This is much simpler then using latest like most puppet modules, and provides a much cleaner way to let you know the file has updated. If you have a newer file, it simply lets you know the checksum has changed.

alexharv074:
Thanks very much for the feedback @nanliu

juniorsysadmin:
@alexharv074 bump

alexharv074:
@juniorsysadmin 

This fell off my radar.

It turned out `s3_checksum` isn't trivial to implement.

I got this:

```
[ec2-user@ip-x-x-x-x ~]$ aws s3api head-object --bucket my-bucket --key some_file.jar
{
    "AcceptRanges": "bytes",
    "ContentType": "application/x-java-archive",
    "LastModified": "Tue, 20 Sep 2016 06:58:57 GMT",
    "ContentLength": 475147311,
    "VersionId": "f1hymsqUPX5m5v5k4JYcwu71lr6X7EhP",
    "ETag": "\"2281d779c3fc85ce73dfdedfa94342aa-8\"",
    "Metadata": {}
}
```

Then I found this page [here](http://stackoverflow.com/questions/1775816/how-to-get-the-md5sum-of-a-file-on-amazons-s3) perhaps hinting how I could then get a checksum from an Etag.

cdenneen:
@nanliu I might be wrong but aren't most implementations of `latest` what cause things like packages and such to become non-idempotent. So in this case, doesn't matter what the checksum is, go grab it again.

I like the idea of the checksum but trying to determine if latest is the proper use case for it.

Part of me thinks `present` is actually one that should be doing the checksum matching to make sure the contents haven't changed upstream... kind of like a file resource with `source => 'puppet:///'`, to enforce state.

If you want something that only downloads once and doesn't "maintain state" then maybe a `only_once => true` option?

cdenneen:
As an aside... shouldn't `creates => '/path/to/non-existent file'` cause this resource to download everytime, essentially getting latest?

ffrank:
No; `creates` is for files that are extracted. It allows Puppet to determine whether an up-to-date archive still needs *extracting* again.

In upstream Puppet, I built the support for updating files from plain HTTP servers using an `mtime` type pseudo-checksum, so that Puppet will download if local `mtime` on disk is older than `LastModified` from the server metadata.
It's not ideal, but it gets the job done in most scenarios.

The ETag does not seem to be a reliable source for an MD5 sum. Trying to use it will probably lead to a wonky UX. Perhaps better to add a "download always" parameter to make just that happen instead, if the user really wants that.

bastelfreak:
Ping @alexharv074 :)

alexharv074:
@bastelfreak , if you want to close / delete this PR feel free.

Issue #225, opened 2016-09-30
1.1.2 does not work with directory environments on Puppetserver 1.1.3 (due to puppet_x)


elconas:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8.7 (Puppetserver 1.1.3-1.el7)
- Ruby: jruby (with puppetserver)
- Distribution: Centos 7 
- Module version: 1.1.2

### How to reproduce (e.g Puppet code you use)

Use environment name "myenvironment".

Puppet run failes with: 

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not autoload puppet/type/archive: Could not autoload puppet/provider/archive/curl: Could not find parent provider ruby of curl on node NODE
```

In this case puppet/archive is consumed by rtyler/jenkins 1.7.0 (https://github.com/jenkinsci/puppet-jenkins/releases/tag/v1.7.0).

### What are you seeing

Class loading failes, caused by https://tickets.puppetlabs.com/browse/SERVER-973 on Puppetserver. 

### What behaviour did you expect instead

Should compile.

elconas:
Maybe you can also do [require File.dirname(__FILE__) + ' /../../path/to/helper/code'](https://github.com/puppetlabs/puppetlabs-mysql/blob/d58a100fa67bc99b4388d4ea3921b11647d483d7/lib/puppet/provider/mysql_database/mysql.rb#L1) as mentioned [here](https://tickets.puppetlabs.com/browse/SERVER-973?focusedCommentId=352751&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-352751)

gregswift:
In the same software environment and same module setup (archive 1.1.2 or 1.1.3-rc0 with jenkinsci-puppet 1.7.0) but with an environment named "jenkins" I saw the same autoload error.  When i researched it i found #105 and made an environment called production. Autoload error stopped and was replaced with

```
Error: Could not autoload puppet/type/archive: /var/lib/puppet/lib/puppet/type/archive.rb:53: syntax error, unexpected ':', expecting ')'
  newparam(:path, namevar: true) do
                          ^
/var/lib/puppet/lib/puppet/type/archive.rb:223: syntax error, unexpected ':', expecting ')'
...aram(:allow_insecure, boolean: true, parent: Puppet::Paramet...
                              ^
/var/lib/puppet/lib/puppet/type/archive.rb:223: Can't assign to true
[...]

PR #226, opened 2016-09-30
Fix https://github.com/voxpupuli/puppet-archive/issues/225

elconas:
Fix Issue https://github.com/voxpupuli/puppet-archive/issues/225

### DIFF
--- a/lib/puppet/parser/functions/artifactory_sha1.rb
+++ b/lib/puppet/parser/functions/artifactory_sha1.rb
@@ -9,7 +9,7 @@ module Puppet::Parser::Functions
     raise(ArgumentError, "Invalid artifactory file info url #{args}") unless args.size == 1
 
     require 'json'
-    require 'puppet_x/bodeco/util.rb'
+    require File.dirname(__FILE__) + '/../../../puppet_x/bodeco/util.rb'
 
     uri = URI(args[0])
     response = PuppetX::Bodeco::Util.content(uri)
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -1,13 +1,5 @@
-begin
-  require 'puppet_x/bodeco/archive'
-  require 'puppet_x/bodeco/util'
-rescue LoadError
-  require 'pathname' # WORK_AROUND #14073 and #7788
-  archive = Puppet::Module.find('archive', Puppet[:environment].to_s)
-  raise(LoadError, "Unable to find archive module in modulepath #{Puppet[:basemodulepath] || Puppet[:modulepath]}") unless archive
-  require File.join archive.path, 'lib/puppet_x/bodeco/archive'
-  require File.join archive.path, 'lib/puppet_x/bodeco/util'
-end
+require File.dirname(__FILE__) + '/../../../puppet_x/bodeco/archive.rb'
+require File.dirname(__FILE__) + '/../../../puppet_x/bodeco/util.rb'
 
 require 'securerandom'
 require 'tempfile'
--- a/spec/unit/puppet_x/bodeco/archive_spec.rb
+++ b/spec/unit/puppet_x/bodeco/archive_spec.rb
@@ -1,6 +1,6 @@
 # rubocop:disable RSpec/MultipleExpectations
 require 'spec_helper'
-require 'puppet_x/bodeco/archive'
+require File.dirname(__FILE__) + '/../../../../lib/puppet_x/bodeco/archive.rb'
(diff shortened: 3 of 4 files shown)


lib/puppet/provider/archive/ruby.rb
```diff
@@ -1,13 +1,5 @@
...
-  require 'puppet_x/bodeco/archive'
-  require 'puppet_x/bodeco/util'
-rescue LoadError
-  require 'pathname' # WORK_AROUND #14073 and #7788
-  archive = Puppet::Module.find('archive', Puppet[:environment].to_s)
-  raise(LoadError, "Unable to find archive module in modulepath #{Puppet[:basemodulepath] || Puppet[:modulepath]}") unless archive
-  require File.join archive.path, 'lib/puppet_x/bodeco/archive'
-  require File.join archive.path, 'lib/puppet_x/bodeco/util'
-end
+require File.dirname(__FILE__) + '/../../../puppet_x/bodeco/archive.rb'
```
alexjfisher: Not sure about this.  The code you're removing looks like it was specifically added as a workaround for https://tickets.puppetlabs.com/browse/SERVER-973

https://github.com/voxpupuli/puppet-corosync/blob/master/lib/puppet/provider/cs_clone/crm.rb#L4 is identical?  But in that module the comments also reference SERVER-973

ping @roidelapluie - Can you clarify?

elconas: I am unsure here. I found [PUB-3502](https://tickets.puppetlabs.com/browse/PUP-3502?focusedCommentId=132688&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-132688) which states:

> "2". The second issue is that your provider is accessing Puppet[:environment] which is almost certainly going to be wrong if you are using more than the production enviroment:
```
     vmware_module = Puppet::Module.find('vmware_lib', Puppet[:environment].to_s)
     require File.join vmware_module.path, "lib/#{path}"`
```
> Both the agent and master will require your vcac_appliance provider, and the use of Puppet[:environment] is likely to be wrong for both of them. For example, the agent setting will default to production but will switch to the ENC specified environment. And the value of the Puppet[:environment] setting on the master will likely not correspond to the agents' environment in which the master is supposed to be compiling a catalog. As a result, if you have different versions of your provider on the master (dev, test, production), then you could be loading helper code from an environment which doesn't match the manifest that you're compiling for the agent.

spec/unit/puppet_x/bodeco/archive_spec.rb
```diff
@@ -1,6 +1,6 @@
 # rubocop:disable RSpec/MultipleExpectations
 require 'spec_helper'
-require 'puppet_x/bodeco/archive'
+require File.dirname(__FILE__) + '/../../../../lib/puppet_x/bodeco/archive.rb'
```
nanliu: There's really no reason to touch unit tests.

lib/puppet/parser/functions/artifactory_sha1.rb
```diff
@@ -9,7 +9,7 @@ module Puppet::Parser::Functions
     raise(ArgumentError, "Invalid artifactory file info url #{args}") unless args.size == 1
 
     require 'json'
-    require 'puppet_x/bodeco/util.rb'
+    require File.dirname(__FILE__) + '/../../../puppet_x/bodeco/util.rb'
```
nanliu: Since functions are pluginsynced to the libdir, this could be loading an arbitrary version of util.rb.

lib/puppet/provider/archive/ruby.rb
nanliu: Since we are ruby 1.9.3+, we can just use require_relative. The patch here might appear to 'fix' the problem, but it's likely coincidental. The core problem is https://tickets.puppetlabs.com/browse/SERVER-94.

olivierHa:
@juniorsysadmin Could you explain why this issue won't be fixed ?

I am seeing other 'supported' modules doing the same thing : 

https://github.com/puppetlabs/puppetlabs-netscaler/blob/master/lib/puppet/provider/netscaler_csvserver/rest.rb#L1

juniorsysadmin:
@olivierHa My understanding after reading the linked issue is that the issue is related to either an EOL Ruby version or Puppet's handing of environments or an older Puppet master version - it's not something that this module should be trying to accommodate. Additionally, the code changes in this PR don't seem to be fixing the underlying issues that were reported. Happy to be corrected and re-open this if it is made clearer the reason for this PR.

PR #237, opened 2016-10-25
Include full Apache License text and add badge

alexjfisher:
github needs the full license text for the detection to work.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,203 @@
-Copyright (c) 2014 Bodeco Inc
+   Copyright 2014 Bodeco Inc
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-    http://www.apache.org/licenses/LICENSE-2.0
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


LICENSE.md
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!) The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2014 Bodeco Inc
```
domcleal: I wouldn't suggest changing this if you're trying to get the licence detection to work, it's part of the appendix of the licence rather than the statement itself. (It's saying: add this text to the source somewhere, not edit this text.)

alexjfisher: I also saw the same thing here https://github.com/voxpupuli/puppet-corosync/blob/master/LICENSE#L189.  But I do see what you mean.  Do you think it's ok for the copyright statement to go in the README instead?

LICENSE
```diff
@@ -1,13 +1,203 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
nanliu: The appendix does not need to be included

nanliu: Ah, the original comment makes much more sense now. Sorry for the confusion.

PR #240, opened 2016-11-02
modulesync 0.15.0

bastelfreak:


dhoppe:
What about the changes regarding simplecov at spec_helper.rb?

PR #245, opened 2016-11-30
refactor and document archive ruby provider

igalic:
use early returns for easier reading of archive's ruby provider.
Since this is the base-provider for wget & curl (and probably future
providers as well), this commit also adds documentation about the
finite-state-machine that we implement (on top of the standard puppet
provider finite-state-machinery).

### DIFF
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -12,22 +12,68 @@ end
 require 'securerandom'
 require 'tempfile'
 
+# This provider implements a simple state-machine. The following attempts to #
+# document it. In general, `def adjective?` implements a [state], while `def
+# verb` implements an {action}.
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/archive/ruby.rb
```diff
@@ -12,22 +12,68 @@
...
+#
+# Here's how we would extend archive for an `ensure => latest`:
+#
+#  [exists?] -> no -> {create}
+#    |
+#    v
+#   yes
+#    |
+#    v
+#  [ttl?] -> expired -> {destroy} -> {create}
```
nanliu: Maybe check Last-Modified?
```
curl -I --header 'If-Modified-Since: Tue, 11 Dec 2012 10:10:24 GMT'
```

igalic: either that, or the `mtime` of the file/archive on the filesystem.

PR #263, opened 2017-02-11
modulesync 0.20.0

bastelfreak:


bastelfreak:
uhm what is broken here?

sacres:
This is odd as `bundle exec rake test` locally doesn't present any failures yet travis does.

PR #264, opened 2017-02-19
Replace validate_* functions with Puppet 4 data type validations

jkroepke:
Added Puppet Data types for every parameter.

remove deprecated validate_* and is_* functions from stdlib.

### DIFF
--- a/manifests/artifactory.pp
+++ b/manifests/artifactory.pp
@@ -11,7 +11,7 @@
 # * url: artifactory download url filepath. NOTE: replaces server, port, url_path parameters.
 # * server: artifactory server name (deprecated).
 # * port: artifactory server port (deprecated).
-# * url_path: artifactory file path http:://{server}:{port}/artifactory/{url_path} (deprecated).
+# * url_path: artifactory file path http://{server}:{port}/artifactory/{url_path} (deprecated).
 # * owner: file owner (see archive params for defaults).
 # * group: file group (see archive params for defaults).
 # * mode: file mode (see archive params for defaults).
@@ -43,20 +43,20 @@
 # }
 #
 define archive::artifactory (
-  $path         = $name,
-  $ensure       = present,
-  $url          = undef,
-  $server       = undef,
-  $port         = undef,
-  $url_path     = undef,
-  $owner        = undef,
-  $group        = undef,
-  $mode         = undef,
-  $archive_path = undef,
-  $extract      = undef,
-  $extract_path = undef,
-  $creates      = undef,
-  $cleanup      = undef,
+  String                                  $path         = $name,
+  Enum['present', 'absent']               $ensure       = present,
+  Optional[Pattern[/^https?:\/\//]]       $url          = undef,
+  Optional[String]                        $server       = undef,
+  Optional[Integer]                       $port         = undef,
+  Optional[String]                        $url_path     = undef,
+  Optional[String]                        $owner        = undef,
+  Optional[String]                        $group        = undef,
+  Optional[String]                        $mode         = undef,
+  Optional[Boolean]                       $extract      = undef,
+  Optional[String]                        $extract_path = undef,
+  Optional[String]                        $creates      = undef,
+  Optional[Boolean]                       $cleanup      = undef,
+  Optional[Stdlib::Compat::Absolute_path] $archive_path = undef,
 ) {
 
   include ::archive::params
@@ -67,7 +67,9 @@ define archive::artifactory (
     $file_path = $path
   }
 
-  validate_absolute_path($file_path)
+  if $file_path !~ Stdlib::Compat::Absolute_path {
+    fail("archive::artifactory[${name}]: \$name or \$archive_path must be an absolute path!") # lint:ignore:trailing_comma
+  }
 
   if $url {
     $file_url = $url
--- a/manifests/go.pp
+++ b/manifests/go.pp
@@ -1,21 +1,21 @@
 # download from go
 define archive::go (
-  $server,
-  $port,
-  $url_path,
-  $md5_url_path,
-  $username,
-  $password,
-  $path         = $name,
-  $owner        = undef,
-  $group        = undef,
-  $mode         = undef,
-  $archive_path = undef,
-  $ensure       = present,
-  $extract      = undef,
-  $extract_path = undef,
-  $creates      = undef,
-  $cleanup      = undef,
+  String                    $server,
+  Integer                   $port,
+  String                    $url_path,
+  String                    $md5_url_path,
+  String                    $username,
+  String                    $password,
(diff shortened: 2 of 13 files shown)


manifests/artifactory.pp
```diff
@@ -67,7 +67,9 @@
     $file_path = $path
   }
 
-  validate_absolute_path($file_path)
+  if $file_path !~ Stdlib::Compat::Absolute_path {
+    fail("archive::artifactory[${name}]: \$name or \$archive_path must be an absolute path!")
```
jkroepke: manifests/artifactory.pp:71:trailing_comma:WARNING:missing trailing comma after last element

manifests/go.pp
```diff
@@ -26,7 +26,9 @@
     $file_path = $path
   }
 
-  validate_absolute_path($file_path)
+  if $file_path !~ Stdlib::Compat::Absolute_path {
+    fail("archive::go[${name}]: \$name or \$archive_path must be an absolute path!")
```
jkroepke: manifests/go.pp:30:trailing_comma:WARNING:missing trailing comma after last element

alexjfisher:
When using `Optional`, do parameters still need to default to `undef`?

jkroepke:
`Optional[X]` is just an alias to `Variant[Undef, X]`.

See: https://docs.puppet.com/puppet/4.9/lang_data_abstract.html#optional

So it's still request to set a default value unless you want to have required parameters.

We can discuss about non undef values for example `cleanup` should be true; https://github.com/voxpupuli/puppet-archive/blob/master/lib/puppet/type/archive.rb#L121 but it should be a own issue/pr.

alexjfisher:
@jkroepke Yeah, ok.  I did a bit more digging.  The reason [this](https://github.com/puppetlabs/puppetlabs-ntp/blob/594b45ec03b0b0eeb6c8e1249bec13d0b24e11a9/manifests/init.pp#L61), (for example), works is that the ntp module also uses 'data in modules' and defaults the parameter to `undef` by using `~` in the yaml [here](https://github.com/puppetlabs/puppetlabs-ntp/blob/594b45ec03b0b0eeb6c8e1249bec13d0b24e11a9/data/common.yaml#L4).

jkroepke:
@alexjfisher correct. But the ntp module used the already deprecated hiera v4 as demo.

Data in modules is imho a task for the puppet 5 age.

alexjfisher:
@jkroepke Good work opening the puppet-lint issue.  Meanwhile, (and so we can get this merged), how about using [control-comments](http://puppet-lint.com/controlcomments/) to disable the check?  You can also include why you've disabled the check.

> For the sake of your memory (and your coworkers), any text in your comment after lint:ignore:<check name> will be considered the reason for ignoring the check and will be displayed when showing ignored problems.

jkroepke:
@alexjfisher Done

jkroepke:
@alexjfisher I have no idea, how to use modern facts on the line like here:

```
Facter.stubs(:value).with(:osfamily).returns 'RedHat'
```

How to modify it to use `facts[os][family]` ?

jkroepke:
@alexjfisher Does it make sense to check failures like

```
 1) archive::nexus nexus archive with allow_insecure => 'foobar' should fail to compile and raise an error matching /"foobar" is not a boolean/
     Failure/Error: it { is_expected.to compile.and_raise_error(%r{"foobar" is not a boolean}) }
       error during compilation: Evaluation Error: Error while evaluating a Resource Statement, Archive::Nexus[/tmp/artifact.war]: parameter 'allow_insecure' expects a Boolean value, got String at line 1 on node testing-docker-98956c12-9e09-4fc4-86af-1202defc7162.ec2.internal
     # ./spec/defines/nexus_spec.rb:150:in `block (3 levels) in <top (required)>'
```

If yes, I haven't the knowlage to adjust this test. Sorry.

alexjfisher:
@jkroepke That test does need to be fixed, but it should be pretty simple. `"foobar" is not a boolean` is the [puppet error](https://github.com/puppetlabs/puppetlabs-stdlib/blob/e64a9d697838d44c01afb5b6cd88c2e3a729f285/lib/puppet/parser/functions/validate_bool.rb#L31) that the `validate_bool` function would have spat out.  As such I think all you need to do is update the regular expression to match the new compilation error output.  ie try replacing this [line](https://github.com/voxpupuli/puppet-archive/blob/9709cf37bf5579dec03ea1721ad184ff6c3657a1/spec/defines/nexus_spec.rb#L150) with...
```ruby
it { is_expected.to compile.and_raise_error(%r{parameter 'allow_insecure' expects a Boolean value, got String}) }
```

alexjfisher:
@jkroepke That looks better.  I don't understand why one of the combinations still failed.  Very odd...  Maybe something to do with the recently introduced parallel rspec??

alexjfisher:
and thanks for the rebase/squash!

alexjfisher:
@jkroepke Other than squashing, did you make any other changes?  Tests are all green this time.

jkroepke:
¯\_(ツ)_/¯

Now its fine.

jkroepke:
> Other than squashing, did you make any other changes? Tests are all green this time.

No, just a squash.

manifests/artifactory.pp
```diff
@@ -43,20 +43,20 @@
...
+  Optional[Integer]                       $port         = undef,
+  Optional[String]                        $url_path     = undef,
+  Optional[String]                        $owner        = undef,
+  Optional[String]                        $group        = undef,
+  Optional[String]                        $mode         = undef,
+  Optional[Boolean]                       $extract      = undef,
+  Optional[String]                        $extract_path = undef,
+  Optional[String]                        $creates      = undef,
+  Optional[Boolean]                       $cleanup      = undef,
+  Optional[Stdlib::Compat::Absolute_path] $archive_path = undef,
```
igalic: how about we use `Tea::Absolutepath` here?

alexjfisher:
I've reworded the PR title as it'll probably get used in the changelog when that's next updated.

manifests/artifactory.pp
jkroepke: hum, why you want to add additional dependencies? The stdlib data types based on the tea types. Not?

DavidS: The ::compat:: types are specifically designed to match the horror that was the validate_ function that went before. There is Stdlib::Absolutepath to do better path checking.

igalic: so we should then, officially, deprecate Tea in favour for Stdlib… not that it was ever in widespread use to begin with…

…

i hope

jkroepke: Improve standards instead build a own world. That why i'm doing this PR. :)

PR #269, opened 2017-03-23
Fixing a typo on credentials file and add the config too

aerostitch:
This PR fixes the name of the credentials file for aws (an "s" is missing) and mention the config file for aws which is generally required as containing the default region.


bastelfreak:
thanks @aerostitch!

PR #270, opened 2017-03-25
Add support for downloading puppet URL’s

hajee:
This PR allows archive to support puppet urls for the source parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,28 +117,16 @@ archive { '/tmp/test100k.db':
 
 ### Puppet URL
 
-Below is an example of how to deploy a tar.gz to a local directory when it is
-served via the ```puppet:///``` style url which is not currently supported.
+Since march 2017, the Archive type also supports puppet url's. Here is an example
+on how to use this:
 
 ```puppet
-$docs_filename = 'help.tar.gz'
-$docs_gz_path  = "/tmp/${docs_filename}"
-$homedir = '/home/myuser/'
 
-# First, deploy the archive to the local filesystem
-file {$docs_gz_path:
-  ensure => file,
-  source => "puppet:///modules/profile/${docs_filename}",
-}
-
-# Then expand the archive where you need it to go
-archive { $docs_gz_path:
-  path          => $docs_gz_path,
-  #cleanup       => true, # Do not use this argument with this workaround for idempotency reasons
+archive { '/home/myuser/help':
+  source        => "puppet:///modules/profile/help.tar.gz',
   extract       => true,
   extract_path  => $homedir,
   creates       => "${homedir}/help" #directory inside tgz
-  require       => [ File[$docs_gz_path] ],
 }
 ```
 
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -122,9 +122,9 @@ Puppet::Type.newtype(:archive) do
   end
 
   newparam(:source) do
-    desc 'archive file source, supports http|https|ftp|file|s3 uri.'
+    desc 'archive file source, supports puppet|http|https|ftp|file|s3 uri.'
     validate do |value|
-      unless value =~ URI.regexp(%w[http https ftp file s3]) || Puppet::Util.absolute_path?(value)
+      unless value =~ URI.regexp(%w[puppet http https ftp file s3]) || Puppet::Util.absolute_path?(value)
         raise ArgumentError, "invalid source url: #{value}"
       end
     end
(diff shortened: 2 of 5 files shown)


bastelfreak:
Hi @hajee, thanks for the great PR!

do you think you can also add some tests?

hajee:
I was looking into that, but decided to first wait for your initial reaction to this PR. What kind of tests would you like, Currently it looks the type has no tests what soever to test for correctness of the specified source.

README.md
```diff
@@ -117,28 +117,16 @@ archive { '/tmp/test100k.db':
 
 ### Puppet URL
 
-Below is an example of how to deploy a tar.gz to a local directory when it is
-served via the ```puppet:///``` style url which is not currently supported.
+Since march 2017, the Archive type also supports puppet url's. Here is an example
```
alexjfisher: I'm guessing that `ruby` is the only provider that will support this feature?  It's not the default provider on Linux.  Does the example need a `provider => ruby`?

lib/puppet/type/archive.rb
```diff
@@ -122,9 +122,9 @@ def should_to_s(value)
   end
 
   newparam(:source) do
-    desc 'archive file source, supports http|https|ftp|file|s3 uri.'
+    desc 'archive file source, supports puppet|http|https|ftp|file|s3 uri.'
     validate do |value|
-      unless value =~ URI.regexp(%w(http https ftp file s3)) || Puppet::Util.absolute_path?(value)
+      unless value =~ URI.regexp(%w(puppet http https ftp file s3)) || Puppet::Util.absolute_path?(value)
         raise ArgumentError, "invalid source url: #{value}"
       end
```
alexjfisher: @hajee Perhaps you could declare a [provider feature](https://docs.puppet.com/guides/custom_types.html#provider-features)?  Something like...
```ruby
feature :puppet_urls, 'The ability to download from Puppet URLs' if value == 'puppet'
```
and then in the ruby provider
```ruby
has_feature :puppet_urls
```

README.md
hajee: I was expecting the default provider to be ruby, but I was wrong. I will update the readme

lib/puppet/type/archive.rb
hajee: Good suggestion. Just wondering: is it possible to do a different validate for the provider too?

README.md
alexjfisher: @hajee Actually, you might be ok.  Ruby isn't the default provider, *but* it is still the parent provider for `curl` and `wget` providers which override the parent's `download` method.  Could you double-check?  For example, looks like things might still go wonky if a user tries to specify a `checksum_url`.

lib/puppet/type/archive.rb
nanliu: Don't need this for now since everything inherits from the ruby provider. `feature :puppet_urls` is only necessary if we start to differentiate features per provider.

README.md
nanliu: So not sure what happens if the puppet:/// url points to a directory v.s. a file. Maybe we should clarify it in documentation about expectations if it doesn't handle directory.

lib/puppet/type/archive.rb
nanliu: Follow up with chat with @alexjfisher, it might be worthwhile to mention in README, puppet:/// source will use ruby provider regardless of provider setting for now. But not a blocker for merging.

nanliu:
Last note, if we want to be smart about conditional downloads, we need to implement checksum retrieval as well via http://www.rubydoc.info/gems/puppet/Puppet/FileServing/Metadata checksum and checksum_type. There isn't a clean way to do this because I'm not aware of way to change default provider based on a parameter. If we could change default provider on the fly based on source url, it will make the implementation much cleaner, instead we need to settle with cramming everything into the ruby provider (instead of ruby/s3/puppet providers).

README.md
hajee: @nanliu : If you mean a puppet url pointing to a directory when using `puppet apply`, it does handle that.

hajee:
Hi guys. Thanks for all the feedback. What are the things that still need attention before this PR can be merged?

alexjfisher:
@hajee Just a note in the README explaining that the ruby provider will be used regardless of provider.  I think we were then good to go.

alexjfisher:
@hajee Could you rebase?

@bastelfreak Are you happy to remove the 'needs-tests' label now?

hajee:
@alexjfisher rebase executed

hajee:
@alexjfisher @bastelfreak Is there an ATA on the merge. I'd hate to do a rebase again :-)

bastelfreak:
Thanks for the PR @hajee!

hajee:
Thanks for all the reviews and suggestions.

PR #257, opened 2017-01-21
Pass proxy values using the wget -e option

walkamongus:
The provider was passing the wget command proxy command line options directly,
but these options do not currently exist. This change passes the proxy
values using the '-e,--execute <command>' option, which executes the given
command as if it were included in a .wgetrc file.

This fixes #256.

PR #272, opened 2017-04-12
Pass proxy values using the wget -e option

nanliu:
The provider was passing the wget command proxy command line options directly,
but these options do not currently exist. This change passes the proxy
values using the '-e,--execute <command>' option, which executes the given
command as if it were included in a .wgetrc file.


nanliu:
Rebase of PR #257 with new rspec tests.

Issue #241, opened 2016-11-04
-z for curl option


athingisathing:
Can we add in a -z option to always download new files if using curl.

athingisathing:
or a way to add in curl params dinamicly

PR #275, opened 2017-04-17
use the default min/max TLS version provided by NSS [RHEL-7]

kalinux:
```RHEL-7 (lib)curl does not enable TLS > 1.0 by default.  Please use the --tlsv1 option of curl to negotiate the highest TLS version supported by client/server.```

Getting error from [puppet-vault](https://forge.puppet.com/jsok/vault/readme) while downloading file .

Found Curl bug in Centos/Redhat 7 [here](https://bugzilla.redhat.com/show_bug.cgi?format=multiple&id=1170339)

Greetings
Karen

### DIFF
--- a/lib/puppet/provider/archive/curl.rb
+++ b/lib/puppet/provider/archive/curl.rb
@@ -19,6 +19,7 @@ Puppet::Type.type(:archive).provide(:curl, parent: :ruby) do
         '-o',
         filepath,
         '-fsSL',
+        '--tlsv1',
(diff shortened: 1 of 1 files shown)


nanliu:
This might make sense as an option per #241, but not as a default.

kalinux:
great ! lets make then an option !

nanliu:
Per discussion, closing in favor of #279 (or some final form of it).

PR #278, opened 2017-05-03
Added support for specifying the region where your S3 bucket lives

TijnBuijs:
Added a new parameter to the Archive module: s3_region. With this optional parameter you can specify in which region your S3 bucket lives. When you don't specify the s3_region, the old behavior is maintained.

When you do specify the s3_region parameter, the --region $s3_region is added to the aws command that this module runs in the end.

If I didn't specify this for my S3 bucket, I got an HTTP 400 response, which prevented me from downloading the file. Now I can download it.

If you have any questions, remarks or improvements, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,17 @@ archive { '/tmp/gravatar.png':
 NOTE: Alternative s3 provider support can be implemented by overriding the
 [s3_download method](lib/puppet/provider/archive/ruby.rb):
 
+If you need to specify custom options to the aws CLI command (eg. the region in
+which your S3 Bucket lives) use the awscli_arguments parameter, which is an
+array of all the options you want, including the needed parameters:
+```puppet
+archive { '/tmp/gravatar.png':
+  ensure           => present,
+  source           => 's3://bodecoio/gravatar.png',
+  awscli_arguments => ['--region', 'eu-central-1']
+}
+```
+
 ### Migrating from puppet-staging
 
 It is recommended to use puppet-archive instead of puppet-staging.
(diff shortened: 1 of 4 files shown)


oranenj:
is `eun-central-1` an aws location?

Can you not use undef instead of :false for the not-defined value?

It'd also be nice if you squashed your commits. 

Other than that, looks good to me. :)

nanliu:
Rather than supporting a specific cli option, it might be better to accept generic s3_options. This seems similar to the need for curl/wget options.

TijnBuijs:
@nanliu, @oranenj: I think I fixed all your comments, except for the squash of the commits. I tried to do that, but because I've pushed them to Github already, this didn't work that well. Sorry for that. I will do that better next time.

But there is a generic option now, and the tests specifies a valid AWS region too (eu-central-1, which is Frankfurt). The code itself doesn't need a specific argument anymore, nor does it assume a specific region while running.

If you have any other remarks, please let me know.

nanliu:
The documentation no longer matches the implementation, see #279 for alternative proposal that will fix both issue. I'm going to leave this open for feedback whether we want per provider parameter or a global parameter.

oranenj:
@TijnBuijs: If you rebase/squash your commits, you'll need to force push to your branch. That'll also update the pull request. Alternatively, this can be merged using the squash and merge option on github, if that's fine with you.

TijnBuijs:
@oranenj I fixed the squashing now. Everything is in one commit now.

@nanliu I fixed the documentation issue. I can live both with my PR being merged or your PR being merged. As long as I can specify in which AWS region the S3 bucket lives, I'm a happy camper :)

roidelapluie:
Has conflicts...

TijnBuijs:
@roidelapluie yeah, I suspect that this has to do with the merge of #279 . Either way, now that #279 is merged, this PR becomes obsolete, because #279 fixes the same issue, but in a more general way. I will close this PR now, because of that.

PR #281, opened 2017-05-08
Implemented ensure => latest

TijnBuijs:
As requested in #223 by @nanliu, I've implemented the s3_checksum function. Can you tell me where to place it and how to get the code to use it?

### DIFF
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -241,7 +241,23 @@ Puppet::Type.type(:archive).provide(:ruby) do
         raise(Puppet::Error, 'HTTP server does not support Last-Modified header, ensure => latest not supported with this HTTP server') if response['Last-Modified'].nil?
         source_mtime = Time.parse(response['Last-Modified']).to_i
         download(archive_filepath) if source_mtime > local_mtime
-      # when %r{^ftp}
+      when %r{^ftp}
+        source_mtime = nil
+        dirpath = File.dirname(resource[:source]) + '/'
+        filename = File.basename(resource[:source])
+        params = curl_params(
+          [
+            dirpath,
+            '-sS'
+          ]
+        )
+        lines = curl(params).split("\n")
+
+        lines.each do |line|
+          lineparts = line.split(' ')
+          source_mtime = Time.parse("#{lineparts[5]} #{lineparts[6]} #{lineparts[7]}").to_i if lineparts.last == filename
+        end
+        download(archive_filepath) if !source_mtime.nil? && source_mtime > local_mtime
       when %r{^file}
         uri = URI(resource[:source])
         path = Puppet::Util.uri_to_path(uri)
@@ -268,8 +284,6 @@ Puppet::Type.type(:archive).provide(:ruby) do
         transfer_download(archive_filepath) if source_mtime > local_mtime
       when nil
         raise(Puppet::Error, 'Unable to fetch archive, the source parameter is nil.')
-      else
-        FileUtils.copy(resource[:source], temppath) unless File.exist?(resource[:source])
       end
     else
       create
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -240,13 +240,8 @@ Puppet::Type.type(:archive).provide(:ruby) do
 
 	  # Get the information from the local disk when the file changed last
       localmodifytime = File.mtime(archive_filepath)
-	  puts s3modifytime.to_i
-	  puts localmodifytime.to_i
 	   if s3modifytime.to_i > localmodifytime.to_i
-         puts "updating"
          transfer_download(archive_filepath)
-	   else
-	     puts "not updating"
 	   end
 	else
       create
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -236,14 +236,14 @@ Puppet::Type.type(:archive).provide(:ruby) do
       ]
       params += resource[:download_options] if resource[:download_options]
       fileinfo = JSON.parse(aws(params))
-	  s3modifytime = Time.parse(fileinfo['LastModified'])
+      s3modifytime = Time.parse(fileinfo['LastModified'])
 
-	  # Get the information from the local disk when the file changed last
+      # Get the information from the local disk when the file changed last
       localmodifytime = File.mtime(archive_filepath)
-	   if s3modifytime.to_i > localmodifytime.to_i
-         transfer_download(archive_filepath)
-	   end
-	else
+      if s3modifytime.to_i > localmodifytime.to_i
+        transfer_download(archive_filepath)
+      end
+    else
       create
     end
   end
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -207,6 +207,18 @@ Puppet::Type.type(:archive).provide(:ruby) do
     )
   end
 
+  def header
+    PuppetX::Bodeco::Util.header(
+      resource[:source],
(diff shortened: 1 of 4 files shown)


nanliu:
ETags isn't directly usable as checksum, see #223 discussion on ETags and following link: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html

> The ETag may or may not be an MD5 digest of the object data. Whether or not it is depends on how the object was created and how it is encrypted as described below

We can make ETag a checksum type, provided we have way to calculate ETag on an existing file and then we can compare S3 ETag metadata v.s. the local file ETag.

TijnBuijs:
@nanliu yeah, the checksum method isn't going to work. I was trying to hook up my new s3_checksum function to the rest of the module, but it went from bad to worse all the time.

I've used a whole different approach now to get the ensure => latest functionality . You can request at S3 when the file was last changed. If this modification time is later then the modification time of the file on disk, the file is downloaded again.
Imho this is a valid way of detecting if the file has been updated or not.

ffrank:
Is the `ruby` provider for S3 only? If it supports other sources, what will happen with `ensure => latests`?

What happens when a different provider is used?

TijnBuijs:
@ffrank yes, you are absolutely correct, in my previous commit only S3 was supported. I'm working on getting the rest of the protocols supported too.
S3, HTTP and local files are supported now. Only FTP is on my ToDo still. Will look into that next.

If you have any further remarks, please let me know.

lib/puppet/provider/archive/ruby.rb
```diff
@@ -219,6 +219,53 @@ def s3_download(path)
...
 
+  def check_latest
+    # Check if the file exists, just download it if not
+    if File.exist?(archive_filepath)
+      localmodifytime = File.mtime(archive_filepath).to_i
+      case resource[:source]
+      when %r{^(http)}
+        uri = URI(resource[:source])
+        response = nil
+        Net::HTTP.start(uri.host, uri.port,
```
nanliu: This needs the same wrapping to handle proxy, ssl validation etc.

lib/puppet/provider/archive/ruby.rb
```diff
@@ -219,6 +219,53 @@ def s3_download(path)
     aws(params)
   end
 
+  def check_latest
+    # Check if the file exists, just download it if not
+    if File.exist?(archive_filepath)
+      localmodifytime = File.mtime(archive_filepath).to_i
```
nanliu: maybe just local_mtime, source_mtime?

lib/puppet/provider/archive/ruby.rb
```diff
@@ -219,6 +219,53 @@ def s3_download(path)
...
+      case resource[:source]
+      when %r{^(http)}
+        uri = URI(resource[:source])
+        response = nil
+        Net::HTTP.start(uri.host, uri.port,
+                        use_ssl: uri.scheme == 'https') { |http| response = http.head(uri.path.to_s) }
+        httpmodifytime = Time.parse(response['Last-Modified']).to_i
```
nanliu: Last-Modified is not guaranteed to be part of the header. I'm not sure what's the fallback.

lib/puppet/provider/archive/ruby.rb
TijnBuijs: I've renamed these variables to local_mtime and source_mtime.

lib/puppet/provider/archive/ruby.rb
TijnBuijs: I've just used the function that is available in PuppetX::Bodeco::Util. This has all the needed checks and functions already. No need for me to re implement what is already there.

lib/puppet/provider/archive/ruby.rb
TijnBuijs: Yeah, me neither. I just let the Puppet run fail now when the Last-Modified header is not in place. If someone wants support for extra headers, this can always be added later.

TijnBuijs:
I think I'm done now with this PR. What is the next step in getting this merged?
Does anyone still has remarks or things they would like to see differently? Then I'm happy to oblige them :)

wottah:
Can anyone give an update on this PR? apart from merge conflicts all seems in order right?

ffrank:
Pinging @nanliu again: Is this fine now?
If nothing happens, please ping me back, I'll merge.

esalberg:
We need this functionality as well.  I'm going to fork the branch for now and rebase, but any idea when this might make it into upstream?

lib/puppet/provider/archive/ruby.rb
igalic: if `Last-Modified` is not present, you essentially have to assume `Last-Modified: now` — especially if the request contained `If-Modified-Since`, and `ETag` is unreliable.

igalic:
i'm okay for merging this as-is, and adding tests by ourselves.
@TijnBuijs would you mind rebasing this?

igalic:
n.b.: i'm able to resolve the conflicts very easily, so if no one is terribly against it, i would merge this (after a release)

igalic:
> @igalic said n.b.: i'm able to resolve the conflicts very easily, so if no one is terribly against it, i would merge this (after a release)

lolololol (i was wrong)

igalic:
i honestly don't know how this happened, https://gist.github.com/e3283d8f8e42fa52d102880acf318232

igalic:
update, here's the same diff, with `-w`: https://gist.github.com/b32a4e98ba38470c8a73587fe1832895
doesn't look half as bad, but i'm still curious as to what caused this

TraGicCode:
> doesn't look half as bad, but i'm still curious as to what caused this

@igalic The spacing/identation was changed.

bastelfreak:
Hi @TijnBuijs, are you still interested in this PR?

carlgarner:
I don't suppose there's any chance of getting this implemented/merged is there?

bastelfreak:
Sadly no, because the tests are failing (and this probably needs a rebase). Are you interested in picking this work up?

carlgarner:
Unfortunately, the incarnation above didn't work as desired, so an alternative workaround was found. We ended up using a Puppet function to grab the ETag value from an S3 object and passing this to the `checksum` option. When these values differ, the function pulls the updated file from the S3 bucket.

Function code used (based on something else found online) was 
```
Puppet::Functions.create_function(:s3_etag) do
    dispatch :etag_lookup do
        param 'String', :s3_bucket
    end

    def etag_lookup(s3_bucket)
        pathparts = s3_bucket.split('/', 4)
        bucket = pathparts[2]
        path = pathparts[3]

        params = [
            '/usr/bin/aws',
            's3api',
            'head-object',
            '--bucket',
            bucket,
            '--key',
            path
        ]

        cmd = params.join(' ')
        output = %x|#{cmd}|
        parsed = JSON.parse(output)

        return parsed['ETag'].to_s.gsub('"', '')
    end
end
```
which allowed us to do this:
```
archive {
  default:
    ensure        => present,
    checksum_type => 'md5',
  ;
  "/etc/ssl/certs/${web_cert}":
    source   => "s3://${cert_bucket}/fullchain.pem",
    checksum => s3_etag("s3://${cert_bucket}/fullchain.pem"),
  ;
  "/etc/ssl/private/${web_key}":
    source   => "s3://${cert_bucket}/privkey.pem",
    checksum => s3_etag("s3://${cert_bucket}/privkey.pem")
  ;
}
```

bastelfreak:
This module or maybe https://github.com/voxpupuli/puppet-extlib (probably extlib) might be a good home for the function. Are you interested to provide it as a pull request? And maybe document the usage in the README.md of with puppet-strings in the function?

bastelfreak:
Hi people,
I'm going to close this PR due to inactivity. Please reopen it if you're still interested.

Issue #282, opened 2017-05-11
Can't download latest SNAPSHOT artifactory artifacts


alexjfisher:
Artifactory Pro users can download 'latest' artifacts. https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API#ArtifactoryRESTAPI-RetrieveLatestArtifact

But it doesn't work with this module. We get a 404 when trying to download the URLs corresponding sha1 using https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API#ArtifactoryRESTAPI-FileInfo in https://github.com/voxpupuli/puppet-archive/blob/6f8db0e467a60d3446dba3897865812c3b8a07a5/lib/puppet/parser/functions/artifactory_sha1.rb#L15

The fileinfo endpoint doesn't do any magic if given a SNAPSHOT url.

PR #284, opened 2017-05-25
Support fetching latest SNAPSHOT artifacts

alexjfisher:
RFC/WIP: DO NOT MERGE

This works for me and fixes
https://github.com/voxpupuli/puppet-archive/issues/282

Feedback and some volunteers to test welcome.

### DIFF
--- a/manifests/artifactory.pp
+++ b/manifests/artifactory.pp
@@ -72,24 +72,43 @@ define archive::artifactory (
   }
 
   if $url {
-    $file_url = $url
-    $sha1_url = regsubst($url, '/artifactory/', '/artifactory/api/storage/')
+    $maven2_data = archive::parse_artifactory_url($url)
+    if $maven2_data and $maven2_data['folder_iteg_rev'] == 'SNAPSHOT'{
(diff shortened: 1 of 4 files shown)


manifests/artifactory.pp
```diff
@@ -72,24 +72,43 @@
   }
 
   if $url {
-    $file_url = $url
-    $sha1_url = regsubst($url, '/artifactory/', '/artifactory/api/storage/')
+    $maven2_data = archive::parse_artifactory_url($url)
```
alexjfisher: if `parse_artifactory_url` fails to parse the url, it returns `nil`

manifests/artifactory.pp
```diff
@@ -72,24 +72,43 @@
...
+    if $maven2_data and $maven2_data['folder_iteg_rev'] == 'SNAPSHOT'{
```
alexjfisher: so if `maven2_data` is nil, the new code won't be run.

bastelfreak:
looks good to me. I would like to merge & release this after a rebase

alexjfisher:
I'd be fine letting it sit in master for a couple of weeks.  There's nothing else waiting for a release I don't think.

PR #285, opened 2017-06-01
Support Nexus 3 urls for artifact downloads

rvdh:
Nexus 3 no longer supports the old url format.

### DIFF
--- a/manifests/nexus.pp
+++ b/manifests/nexus.pp
@@ -26,6 +26,7 @@ define archive::nexus (
   Enum['none', 'md5', 'sha1', 'sha2','sh256', 'sha384', 'sha512'] $checksum_type   = 'md5',
   Boolean           $checksum_verify = true,
   String            $packaging       = 'jar',
+  Boolean           $use_nexus3_urls = false,
   Optional[String]  $classifier      = undef,
   Optional[String]  $extension       = undef,
   Optional[String]  $username        = undef,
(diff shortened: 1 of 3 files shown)


rvdh:
@alexjfisher @nanliu Can someone look at this PR? 😄

manifests/nexus.pp
```diff
@@ -65,9 +66,20 @@
...
-  $artifact_url = assemble_nexus_url($url, $query_params)
-  $checksum_url = regsubst($artifact_url, "p=${packaging}", "p=${packaging}.${checksum_type}")
-
+  if $use_nexus3_urls {
+    if $classifier {
+      $c = "-${classifier}"
+    } else {
+      $c = ''
+    }
+    $artifact_url = sprintf('%s/repository/%s/%s/%s/%s/%s-%s%s.%s', $url,
```
alexjfisher: Can any of the parameters, (artifact_id, version etc), contain characters that would need escaping when building a URL?

rvdh: They can (but shouldn't). It would be very unusual to have, say, a space in one of the parameters.

alexjfisher: or a slash?

rvdh: A slash should never be in any of the parameters. Either way, if there is a slash and it's escaped you'd get a 404 and if it's not escaped, you'd also get a 404. 

But if you're being silly and you specify the group_id with slashes (nl/avisi/foo instead of nl.avisi.foo) - escaping the slash will actually break the url.

manifests/nexus.pp
```diff
@@ -43,6 +43,7 @@
   Optional[String]  $proxy_server    = undef,
   Optional[String]  $proxy_type      = undef,
   Optional[Boolean] $allow_insecure  = undef,
+  Optional[Boolean] $use_nexus3_urls = false,
```
alexjfisher: No need for 'Optional' here as you're not defaulting to 'undef'

alexjfisher:
@rvdh Thanks for making that change.  Could you rebase/squash down to a single commit?

rvdh:
@alexjfisher Ofcourse, done! Thanks for taking the time to review.

zipkid:
Hi, 

Is there a reason this is not merged yet?

alexjfisher:
@zipkid Thanks.

Issue #289, opened 2017-07-25
HTTPS download broken again on windows


krjensen:
Hi, 

I try to use archive to download Consul from: https://releases.hashicorp.com/consul/0.9.0/consul_0.9.0_windows_amd64.zip

But it fails with
```
Error: Could not set 'present' on ensure: SSL_connect returned=1 errno=0 state=e
rror: certificate verify failed at 50:/etc/puppetlabs/code/consul
/manifests/install.pp
```

I can find an old issue on this, but it was fixed. 
Im using windows 2016

krjensen:
Adding `allow_insecure: true` works, but it's not a pretty solution.

ElvenSpellmaker:
I'm getting this on using _http_ which is even more odd...

`source          => "http://download.oracle.com/otn-pub/java/jce/8/${jce_filename}",`

```
Error: Could not set 'present' on ensure: SSL_connect returned=1 errno=0 state=S
SLv3 read server certificate B: certificate verify failed at 46:/etc/puppet/modu
les/profile/manifests/configurations/java/jce.pp
Error: Could not set 'present' on ensure: SSL_connect returned=1 errno=0 state=S
SLv3 read server certificate B: certificate verify failed at 46:/etc/puppet/modu
les/profile/manifests/configurations/java/jce.pp
Wrapped exception:
SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificat
e verify failed
Error: /Stage[main]/Profile::Configurations::Java::Jce/Archive[C:\Temp\jce_polic
y-8.zip]/ensure: change from absent to present failed: Could not set 'present' o
n ensure: SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B:
certificate verify failed at 46:/etc/puppet/modules/profile/manifests/configurat
ions/java/jce.pp
```

nanliu:
probably due to the age of the cacert.pem used to gloss over an [old issue](https://github.com/voxpupuli/puppet-archive/issues/188#issuecomment-233668992) for windows. someone should spend some time to see if this no longer needed. 

in the mean time a newer cacert.pem file is likely to fix this: https://curl.haxx.se/docs/caextract.html

ElvenSpellmaker:
@nanilu How would this affect an http link though? It shouldn't be using certs for http?

nanliu:
@ElvenSpellmaker probably due to redirect to https.

ElvenSpellmaker:
[...]

PR #290, opened 2017-08-02
Update cacert.pem

nanliu:


nanliu:
Might fix #289. Provided for convenience and feedback, not tested.

Issue #199, opened 2016-08-17
archive_windir fact doesn't work on Ruby 1.8.7


treydock:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8.6
- Ruby: 1.8.7
- Distribution: RHEL6
- Module version: 1.0.0

### What are you seeing

Facter syntax errors

### What behaviour did you expect instead

No errors

### Output log

```
Error loading fact /var/lib/puppet/lib/facter/archive_windir.rb: /var/lib/puppet/lib/facter/archive_windir.rb:2: syntax error, unexpected ':', expecting kEND
  confine osfamily: :windows
```

### Any additional information you'd like to impart

The hash style used in the archive_windir fact does not work in Ruby 1.8.7 which is used by RedHat 6.

nanliu:
The latest version of this module does not support ruby 1.8.7. See release note for deprecation info.

Issue #152, opened 2016-03-31
Method HTTP.follow_redirect works only with Ruby 1.9+


thiagomarinho:
In file lib/puppet_x/bodeco/util.rb, method HTTP.follow_redirect contains this code:

```
Net::HTTP.start(uri.host, uri.port, :use_ssl => (uri.scheme == 'https')) do |http|
# ...
```

But this syntax (keyword arguments) is supported only by Ruby 2.x, right? Also, I think this option (use_ssl) doesn't exist in previous versions of the class Net::HTTP.start (as you can see here: http://ruby-doc.org/stdlib-1.8.7/libdoc/net/http/rdoc/Net/HTTP.html)

I tested without :use_ssl parameter and it worked well for me (I tested with archive::nexus, using http, and archive, using https), but I don't know if this change could impact.

jyaworski:
@thiagomarinho we no longer 'officially' support Ruby 1.8.7. However, we do still support 1.9.3. If this only works in 2+, we should take a look.

thiagomarinho:
@jyaworski Thanks for your answer. I can't test right now with 1.9.3 since puppet seems to not work with rvm, but I can do it later and report here. But I think that the problem also occurs with this version because keyword arguments is not available before 2.x

jyaworski:
Tagging @nanliu

thiagomarinho:
Hey @jyaworski , sorry for the delay. I tested with Ruby 1.9.3 and 2.0.0 and it worked with both versions. So, as archive doesn't support 1.8.7 anymore, I think that this issue can be closed, do you agree? :(

thiagomarinho:
Maybe we could make the code compatible with 1.8.7 too, what do you think?

bastelfreak:
:-1: I'm against adding support for 1.8.7, this is a really really old ruby version which is EOL. Please upgrade your ruby instead of adding 1.8.7 support in that module.

igalic:
supporting 1.8 is a *lot* of work.
if you can provide a (monkey?) patch that works for people still stuck on ruby 1.8, we'll most likely merge it. and then three months down the road someone will add a feature, or fix a bug, and it will fall over again.

let me backtrack here and explain ***why*** supporting 1.8 — for us — is a lot of work:

[...]

PR #292, opened 2017-08-09
Ruby 1.8 support

sathieu:
Ruby 1.8 doesn't support this hash syntax.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -41,6 +41,7 @@ Lint/LiteralInInterpolation:
 
 Style/HashSyntax:
   Enabled: True
+  EnforcedStyle: 'hash_rockets'
 
 Style/RedundantReturn:
   Enabled: True
--- a/lib/facter/archive_windir.rb
+++ b/lib/facter/archive_windir.rb
@@ -1,5 +1,5 @@
 Facter.add(:archive_windir) do
-  confine :osfamily => :windows # rubocop:disable Style/HashSyntax
+  confine :osfamily => :windows
   setcode do
     program_data = `echo %SYSTEMDRIVE%\\ProgramData`.chomp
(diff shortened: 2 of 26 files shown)


alexjfisher:
@sathieu Ruby 1.8 hasn't been supported in Vox Pupuli maintained modules for a very long time now.

Puppet 3 support was also dropped at the beginning of the year when it became End Of Life.  Puppet 4 (and 5) ships with its own ruby, so even if you're using CentOS/Redhat 6 (which uses ruby 1.8 as its system ruby), you should be fine using this module.

alexjfisher:
For further explanation of why ruby 1.8 isn't supported, please see https://github.com/voxpupuli/puppet-archive/issues/199 and https://github.com/voxpupuli/puppet-archive/issues/152

@sathieu Sorry if this isn't the response you were hoping for.  If you need help upgrading to puppet 4/5, please feel free to ask on the #voxpupuli IRC channel.  There are plenty of people there who would be more than happy to provide assistance.

sathieu:
OK. this is not a surprise.

Note that my puppetmasters are with Puppet 4, but I have some old agents still alive. i.e, I need ruby1.8 support only client-side.

alexjfisher:
IIRC I think it's actually more than just hash syntax in this module.  Something in https://github.com/voxpupuli/puppet-archive/commit/fc80536099e6ec471a79b64d4453212873be052d didn't work properly in ruby 1.8.  Can't remember what it was though!

sathieu:
@alexjfisher Wasn't it 0ef5362f270ee308cf790e1c1878479e3466321b ? Because ''.to_sym raise "ArgumentError: interning empty string" on Ruby 1.8.

Would you merge my changes if I fix this too?

PR #295, opened 2017-08-14
Enable allow_insecure in archive::download

### LINKED ISSUE #294: allow_insecure is not working
I am trying to download a file to a specific location(no compress/decompress)

```ruby
 archive::download { 'gitlab-ci-multi-runner.rpm':
    url            => "https://myserver/gitlab-ci-multi-runner.rpm",
    allow_insecure => true,
  }
```

```
(60) Peer's certificate issuer has been marked as not trusted by the user.
==> : More details here: http://curl.haxx.se/docs/sslcerts.html
==> : 
==> : curl performs SSL certificate verification by default, using a "bundle"
==> :  of Certificate Authority (CA) public keys (CA certs). If the default
==> :  bundle file isn't adequate, you can specify an alternate file
==> :  using the --cacert option.
==> : If this HTTPS server uses a certificate signed by a CA represented in
==> :  the bundle, the certificate verification probably failed due to a
==> :  problem with the certificate (it might be expired, or the name might
==> :  not match the domain name in the URL).
==> : If you'd like to turn off curl's verification of the certificate, use
==> :  the -k (or --insecure) option.
```

alexjfisher:
When the camptocamp archive::download class was first added,
allow_insecure wasn't a feature present in puppet/archive.

Fixes #294

### DIFF
--- a/manifests/download.pp
+++ b/manifests/download.pp
@@ -38,7 +38,7 @@ define archive::download (
   Optional[Enum['none', 'md5', 'sha1', 'sha2','sh256', 'sha384', 'sha512']] $digest_type      = 'md5',   # bad default!
   Integer                       $timeout          = 120,     # ignored
   Stdlib::Compat::Absolute_path $src_target       = '/usr/src',
-  Boolean                       $allow_insecure   = false,   # ignored
+  Boolean                       $allow_insecure   = false,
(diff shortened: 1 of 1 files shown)


igalic:
yes.

PR #297, opened 2017-08-25
Release 2.0.0

alexjfisher:
Updated metadata to mark module as puppet 5 compatible.
Moved to using github_changelog_generator with old CHANGELOG moved to
HISTORY.md.


HISTORY.md
```diff
@@ -0,0 +1,198 @@
+## v1.2.0 (2016-12-25)
```
alexjfisher: HISTORY.md is basically a copy of what was CHANGELOG.md

The release headings have been tweaked so that they are consistent with the autogenerated ones and I've removed the v1.3.0 entry so that github_changelog_generator automatically uses v1.2.0 as the `--since-tag`.  Using v1.3.0 didn't work properly.  I think this is a bug in the tool.  (--since-tag can't be the latest tag)

hajee:
@alexjfisher Thanks for all the work!

Issue #105, opened 2015-12-23
 autoload puppet/provider/archive/curl


eperdeme:
Hello,

I get this error randomly on my clients and if I re-run puppet sometimes it goes away and works without any problems. Any idea why it randomly occurs ?

Puppet: 3.8.2
Module; 0.4.4

Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Could not autoload puppet/type/archive: Could not autoload puppet/provider/archive/curl: Could not find parent provider ruby of curl at /etc/puppetlabs/code/environments/develop/modules/archive/manifests/nexus.pp:53:3 at /etc/puppetlabs/code/environments/develop/modules/ui_broker/manifests/init.pp:33 on node rnode01.test.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run

igalic:
that's interesting / bizarre.

daenney:
Alright, breaking down the issue:

It's complaining that:
> Could not autoload puppet/provider/archive/curl: Could not find parent provider ruby of curl

This relates to [L1 of `lib/provider/puppet/archive/curl.rb`](https://github.com/voxpupuli/puppet-archive/blob/ed94e21bc1bd53846572b6e523cbb2b8e3e4ce36/lib/puppet/provider/archive/curl.rb#L1) which does:

```ruby
Puppet::Type.type(:archive).provide(:curl, :parent => :ruby) do
```

The `ruby` parent provider we can see [here](https://github.com/voxpupuli/puppet-archive/blob/ed94e21bc1bd53846572b6e523cbb2b8e3e4ce36/lib/puppet/provider/archive/ruby.rb).

As long as those two files are in the right location on your Puppet master and plugin sync is turned on everywhere this Should Work :tm:.

Project0:
i have exactly the same issue with wget, but it occures every time and i dont know how to fix. Pluginsync works fine and all files are present on destination node. 
`Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not autoload puppet/type/archive: Could not autoload puppet/provider/archive/wget: Could not find parent provider ruby of wget' 
Puppet 3.6.1
Module 0.4.4

Project0:
I solved the problem right now, it seems to be a problem on the server side when using environments. 
[...]

Issue #248, opened 2016-12-05
archive not found with puppet apply


mattford63:
puppet-archive seems to ignore the --modulepath setting on execution (at least for me).

```
Debug: Caching environment 'production' (ttl = 0 sec)
Error: Could not autoload puppet/provider/archive/ruby: Unable to find archive module in modulepath /etc/puppet/modules:/opt/puppetlabs/puppet/modules
Error: Could not autoload puppet/type/archive: Could not autoload puppet/provider/archive/ruby: Unable to find archive module in modulepath /etc/puppet/modules:/opt/puppetlabs/puppet/modules
Error: Evaluation Error: Error while evaluating a Resource Statement, Could not autoload puppet/type/archive: Could not autoload puppet/provider/archive/ruby: Unable to find archive module in modulepath /etc/puppet/modules:/opt/puppetlabs/puppet/modules at /home/matt/src/bitbucket/workstation/puppet/modules/terraform/manifests/init.pp:27:3 on node gamingpc.lan
```

The run command looks like this

```
puppet apply --hiera_config=/home/matt/src/bitbucket/workstation/puppet/hiera.yaml --modulepath=/home/matt/src/bitbucket/workstation/puppet/modules:/home/matt/src/bitbucket/workstation/puppet/localmodules /home/matt/src/bitbucket/workstation/puppet/manifests/site.pp --color=true --verbose --debug --show_diff --write-catalog-summary
~/src/bitbucket/workstation
```

mattford63:
After playing around with some different puppet installs it seems as though it defaults to the original search locations.

ffrank:
Hi, can you provide a simple reconstruction case?

Is `archive` behaving differently than other type/provider modules in this context?

kwisatz:
#105 and #225 might be related, though they're in a master-server context, not using puppet apply.

ffrank:
Interesting, although those feel different.

What we'd really need here is a step-by-step how to create an environment that exhibits this specific issue.

baurmatt:
Having the same issues here:

```bash
mkdir modules
git clone https://github.com/voxpupuli/puppet-archive modules/archive
cat >> test.pp >> EOF
archive { '/home/mbaur/g10k':
  ensure  => present,
  source  => 'https://github.com/xorpaul/g10k/releases/download/v0.3.14/g10k-linux-amd64.zip',
  extract => true,
}
EOF
puppet apply --modulepath=modules/ test.pp
```

[...]

PR #307, opened 2017-11-23
Fix puppetmaster issue with parent loading in multiple environments

hajee:
After a restart, the puppetmaster was (sometimes) unable to load the child providers because the parent provider was unknown. This commit fixes this issue.

### DIFF
--- a/lib/puppet/provider/archive/curl.rb
+++ b/lib/puppet/provider/archive/curl.rb
@@ -1,4 +1,6 @@
-Puppet::Type.type(:archive).provide(:curl, parent: :ruby) do
+require_relative './ruby'
+
+Puppet::Type.type(:archive).provide(:curl, parent: Puppet::Type::Archive::ProviderRuby) do
   commands curl: 'curl'
   defaultfor feature: :posix
 
--- a/lib/puppet/provider/archive/wget.rb
+++ b/lib/puppet/provider/archive/wget.rb
@@ -1,4 +1,6 @@
-Puppet::Type.type(:archive).provide(:wget, parent: :ruby) do
+require_relative './ruby'
+
+Puppet::Type.type(:archive).provide(:wget, parent: Puppet::Type::Archive::ProviderRuby) do
   commands wget: 'wget'
 
   def wget_params(params)


alexjfisher:
Was it just the lack of `require`, or the way the parent provider was declared?
`parent: Puppet::Type::Archive::ProviderRuby` looks odd (but might be correct).  The docs suggest the original way was correct though.
https://puppet.com/docs/puppet/5.3/provider_development.html#another-provider-of-the-same-resource-type

alexjfisher:
I had a similar issue a couple of days ago after restarting a puppetserver.  It worked *after* the agent on the same node had run puppet.  I've got a hectic day ahead of me tomorrow, but I'd like to see if I can reproduce the problem on Monday.

@hajee Could you provide more details about your environment, the exact error message, Puppet and puppetserver versions etc.  What have you got [ruby-load-path](https://github.com/puppetlabs/puppetserver/blob/961e889bc370890d6ee0be9e9a6ffbd4e35a372d/resources/ext/config/conf.d/puppetserver.conf#L5) set to?  (I think lots of people also include their agent's path as a workaround to various issues).  Probably not applicable, but do you use `puppet generate types`?

This is such a widely used module, I personally err on the side of extreme caution with this sort of change.  A fix for one set of users might break it for another etc.

hajee:
The error is:

```
Could not find parent provider ruby of wget
```
## How to reproduce

We can reproduce this error by making sure the module is **NOT** installed in a production environment and **IS** installed in an other environment. Then stop and start the puppetserver and run then run the agent.  Only on the first run of the agent, this error is generated. When you run the agent again. It works just fine and keeps on running just fine.

## Analysis

The original error is generated inside the [`self.provide` method from `Puppet::Type`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type.rb#L1827). This method supports two ways of specifying the type. One where the name is a string or a symbol. This is syntactic sugar. [In this case](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type.rb#L1839) Puppet does some mambojambo to get the real ruby class name of the parent. The error is generated when puppet tries to get the class name from the string. I'm not sure how and why, but the solution I choose is to use the other code path.

[The other option](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type.rb#L1836) is to  specify a **real** class (The class constant name). In that case this is just used as the parent.

## Solution

The `require_relative` for the parent class file, makes sure that on ruby level the class is defined and available. So no need for Puppet to some fancy lading. Because we require the parent class from the **same path** as the current class, we are sure it the current class and the parent class ware always loaded from the same Puppet environment. 

I'm pretty sure there is no requirement for changing ruby load paths. At least we didn't do this to get it working.

hajee:
@alexjfisher Did you manage to squeeze in a test in the weekend? I tried to make a clear description of the issue, how to reproduce this and the solution.

alexjfisher:
@hajee Sorry, not yet. But thanks for the very detailed description and analysis. It really helps.
@ekohl Any further thoughts?

alexjfisher:
For what it's worth, I found extra discussion [here](https://tickets.puppetlabs.com/browse/SERVER-94?focusedCommentId=301620&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-301620).

alexjfisher:
Looks like this would fix https://github.com/voxpupuli/puppet-archive/issues/105 and maybe https://github.com/voxpupuli/puppet-archive/issues/248 ?

binford2k:
Does `puppet generate types` resolve the issue without using this patch?

hajee:
@binford2k: I just tried after using `puppet generate types`, but unfortunately it doesn't make a difference. 

@alexjfisher:I think this PR will close #105. But I'm not sure about #248.

I think indeed issue [SERVER-94](https://tickets.puppetlabs.com/browse/SERVER-94) might be connected. Although a lot has already been done on environment isolation, our issue is still unfixed since 2014/10/21. Personally I think this PR might not provide the fanciest code to look at, but it does work around the core issue in Puppet. I'm looking forward to other views on this :-)

I hope anyone can find the time to test this and merge it.

hajee:
Is there any progress in reviewing this PR?

alexjfisher:
It's taken me a while this morning to reproduce the problem.

- I removed the archive module from my puppetserver's environment
- Removed `/opt/puppetlabs/puppet/cache/lib` from `ruby-load-path` in puppetserver.conf (This is needed for me to workaround issues with loading functions).

After restarting the puppetserver I *still* couldn't reproduce the problem until I deleted `/etc/puppetlabs/code/environments/the_agent_environment_with_archive_in/.resource_types`

Then after restarting the puppetserver, the agent run fails with `Could not find parent provider ruby of curl`.  Not just on the first run, but every single time, (I had no other agents using this server, so that could possibly explain this?)

Running `umask 022; puppet generate types --environment=the_agent_environment_with_archive_in` fixed it without needing to restart the puppetserver.

@hajee Could you double check `generate types` really doesn't fix it for you?  (At first, I forgot my system umask was 027 and when I ran `generate types` as root, the files created weren't readable by the puppetserver user).

hajee:
@alexjfisher was right on the spot. The `puppet generate types` **WITH** the correct `umask` does fix the issue.  Thanks for the hard work.

I'll close this issue. For users that are not or not yet using a Puppet server that had type generation, they can use this fix still.

Xtigyro:
Puppet Could not autoload puppet/provider/archive/ruby: Unable to find archive module in ```
modulepath /etc/puppetlabs/code/modules:/opt/puppetlabs/puppet/modules
/etc/puppetlabs/code/environments/dev/modules_forge/archive/lib/puppet/provider/archive/ruby.rb:7:in `<main>'
org/jruby/RubyKernel.java:1022:in `load'
```
@hajee @alexjfisher 
I'm having a similar error - using Puppet Server `v6.7.1` in Docker (Kubernetes).
Could we reopen this issue please?

alexjfisher:
@Xtigyro Are you running `puppet generate types` after deploying/updating your environment?

Xtigyro:
> 
> 
> @Xtigyro Are you running `puppet generate types` after deploying/updating your environment?

@alexjfisher 
No but why do I need to do that? I'm using the latest version of Puppet.

PR #318, opened 2018-03-07
Don't glob archive URL with curl

derekhiggins:
Allows IPv6 IP addresses to be used wraped in square
brackets.

See https://bugs.launchpad.net/tripleo/+bug/1754065


EmilienM:
Thanks!

Issue #321, opened 2018-03-12
Convert functions to puppet 4 API


alexjfisher:
Converting the puppet 3 API functions to use the Puppet 4 API, should fix environment isolation issues people occasionally have.

For example, it will probably fix https://github.com/voxpupuli/puppet-archive/issues/320

trevor-vaughan:
FYI, the way the `include` statement works in `go_md5` causes compilation to fail on occasion. It either needs to be ported to Puppet 4 or use a relative include.

alexjfisher:
> FYI, the way the `include` statement works in `go_md5` causes compilation to fail on occasion. It either needs to be ported to Puppet 4 or use a relative include.

Are *you* a GoCD user? See https://github.com/voxpupuli/puppet-archive/pull/392

PR #322, opened 2018-03-12
Remove deprecated archive::artifactory parameters

alexjfisher:
The following parameters were deprecated a long time ago and have now
been removed.

* `server`
* `port`
* `url_path`

Removing them is a breaking change but hopefully most users are already
using `url`

### DIFF
--- a/manifests/artifactory.pp
+++ b/manifests/artifactory.pp
@@ -8,10 +8,7 @@
 #
 # * path: fully qualified filepath for the download the file or use archive_path and only supply filename. (namevar).
 # * ensure: ensure the file is present/absent.
-# * url: artifactory download url filepath. NOTE: replaces server, port, url_path parameters.
-# * server: artifactory server name (deprecated).
-# * port: artifactory server port (deprecated).
-# * url_path: artifactory file path http://{server}:{port}/artifactory/{url_path} (deprecated).
+# * url: artifactory download URL.
 # * owner: file owner (see archive params for defaults).
 # * group: file group (see archive params for defaults).
 # * mode: file mode (see archive params for defaults).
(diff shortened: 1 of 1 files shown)


alexjfisher:
I would like to get this in before [converting the artifactory_sha1](https://github.com/voxpupuli/puppet-archive/issues/321) function to the puppet 4 API.  As removing the deprecated parameters is already a breaking change, the function migration will probably be too (eg. I'll probably namespace it)

manifests/artifactory.pp
```diff
@@ -43,12 +40,9 @@
 # }
 #
 define archive::artifactory (
+  Pattern[/^https?:\/\//]                 $url,
```
ekohl: Sounds like ```Stdlib::Httpurl```

manifests/artifactory.pp
```diff
@@ -9,9 +9,6 @@
 # * path: fully qualified filepath for the download the file or use archive_path and only supply filename. (namevar).
 # * ensure: ensure the file is present/absent.
 # * url: artifactory download url filepath. NOTE: replaces server, port, url_path parameters.
```
ekohl: Is the note still relevant? Maybe better to do that in the changelog?

manifests/artifactory.pp
```diff
@@ -71,44 +65,29 @@
     fail("archive::artifactory[${name}]: \$name or \$archive_path must be an absolute path!") # lint:ignore:trailing_comma
```
ekohl: Odd lint ignore here. Also wondering if we should move the above line to just ```Stdlib::Absolutepath```

alexjfisher: I tried removing the lint, and puppet-lint *does* complain.  A false positive I guess.
Maybe I should use `assert_type` instead.

manifests/artifactory.pp
alexjfisher: Fixed

manifests/artifactory.pp
alexjfisher: Fixed.

manifests/artifactory.pp
alexjfisher: I've updated this code to use assert_type and Stdlib::Absolutepath

PR #323, opened 2018-03-12
Rewrite artifactory_sha1 function with puppet v4 api

alexjfisher:
Old v3 API function is removed, and new `archive::artifactory_checksum`
function added.  Puppet 4 API functions should not have issues with
environment isolation.

This is a breaking change.  The function is renamed and has been
upgraded. It can now return md5 or sha256 checksums as well as sha1.

The function now expects the artifact URL and will convert to the API
url internally. (This was previously done in the calling puppet code.)


alexjfisher:
@eputnam puppet-strings doesn't seem very happy with me.

```
puppet strings generate --format markdown
--- SNIP ---
[warn]: Missing documentation for Puppet function 'archive::artifactory_checksum' at lib/puppet/functions/archive/artifactory_checksum.rb:4.
[error]: Unhandled exception in PuppetStrings::Yard::Handlers::Ruby::FunctionHandler:
  in `lib/puppet/functions/archive/artifactory_checksum.rb`:4:

        4: Puppet::Functions.create_function(:'archive::artifactory_checksum') do

[error]: NoMethodError: undefined method `types=' for nil:NilClass
[error]: Stack trace:
        /home/alex/.rvm/gems/ruby-2.4.1/gems/puppet-strings-1.2.1/lib/puppet-strings/yard/handlers/ruby/function_handler.rb:142:in `block in add_overload_tag'
        /home/alex/.rvm/gems/ruby-2.4.1/gems/puppet-strings-1.2.1/lib/puppet-strings/yard/handlers/ruby/function_handler.rb:135:in `each'
        /home/alex/.rvm/gems/ruby-2.4.1/gems/puppet-strings-1.2.1/lib/puppet-strings/yard/handlers/ruby/function_handler.rb:135:in `add_overload_tag'
        /home/alex/.rvm/gems/ruby-2.4.1/gems/puppet-strings-1.2.1/lib/puppet-strings/yard/handlers/ruby/function_handler.rb:100:in `block in add_tags'
        /home/alex/.rvm/gems/ruby-2.4.1/gems/puppet-strings-1.2.1/lib/puppet-strings/yard/handlers/ruby/function_handler.rb:98:in `each'
        /home/alex/.rvm/gems/ruby-2.4.1/gems/puppet-strings-1.2.1/lib/puppet-strings/yard/handlers/ruby/function_handler.rb:98:in `add_tags'
```

ekohl:
@alexjfisher should the docs be above the function instead of inside it perhaps? Now it looks like it's documenting the dispatch.

eputnam:
having them above the dispatch is correct. i'm going to take a closer look.

alexjfisher:
I think it's the `return_type` it doesn't like.

eputnam:
looks like this is another bug. if `return_type` is used, strings assumes a `@return` tag is present. if the return tag is not present (or nil), it'll throw this error. if you just add an empty return tag, you can get around this.

lib/puppet/functions/archive/artifactory_checksum.rb
```diff
@@ -0,0 +1,25 @@
...
+
+Puppet::Functions.create_function(:'archive::artifactory_checksum') do
+  # @summary A function that returns the checksum value of an artifact stored in Artifactory
+  # @param url The URL of the artifact.
+  # @param checksum_type The checksum type.
+  #        Note the function will raise an error if you ask for sha256 but your artifactory instance doesn't have the sha256 value calculated.
+  dispatch :artifactory_checksum do
+    param 'Stdlib::HTTPUrl', :url
+    optional_param "Enum['sha1','sha256','md5']", :checksum_type
+    return_type 'String'
```
alexjfisher: If I remove this line, the docs generate.
Equally, I can also add a `# @return [String] The checksum` above.
I don't think this should be required though as it doesn't add much if I'm already doing the `return_type` validation thing.

eputnam: https://tickets.puppetlabs.com/browse/PDOC-229

spec/defines/artifactory_spec.rb
```diff
@@ -3,10 +3,9 @@
 describe 'archive::artifactory' do
   let(:facts) { { os: { family: 'RedHat' }, puppetversion: '4.4.0' } }
 
-  before do
-    MockFunction.new('artifactory_sha1') do |f|
-      f.stub.returns('0d4f4b4b039c10917cfc49f6f6be71e4')
-    end
+  # Mock Puppet V4 API ruby function with a puppet language function equivalent
+  let(:pre_condition) do
+    'function archive::artifactory_checksum($url,$type) { return \'0d4f4b4b039c10917cfc49f6f6be71e4\' }'
```
alexjfisher: Many thanks to @TraGicCode on slack for suggesting this method of mocking puppet v4 API ruby functions with puppet language functions.

PR #349, opened 2018-08-13
Put the cookie option at the end when using curl

kapouik:
When using curl to download, the --cookie option is currently put in first before --proxy or --insecure options. Sometimes, the content of cookie can broke all command put after (discover with the module puppetlabs/puppetlabs-java when you use the class java::oracle).
So I just change the options order to make it work.

No issue open

### DIFF
--- a/lib/puppet/provider/archive/curl.rb
+++ b/lib/puppet/provider/archive/curl.rb
@@ -5,10 +5,10 @@ Puppet::Type.type(:archive).provide(:curl, parent: :ruby) do
   def curl_params(params)
     account = [resource[:username], resource[:password]].compact.join(':') if resource[:username]
     params += optional_switch(account, ['--user', '%s'])
-    params += optional_switch(resource[:cookie], ['--cookie', '%s'])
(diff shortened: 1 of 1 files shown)


ekohl:
The order shouldn't matter so I'm interested in how can it exactly breaks. Insufficient shell escaping? I wonder if that could be a remote root exploit since the remote has control over cookies.

kapouik:
You can try this command : 
curl http://download.oracle.com/otn-pub/java/jdk//7u80-b15/jdk-7u80-linux-x64.rpm -o /tmp/jdk-7u80-linux-x64.rpm_20180810-6773-1a6m96j -fsSL --max-redirs 5 --cookie gpw_e24=http%3A%2F%2Fwww.oracle.com%2F; --proxy http://127.0.0.1:3127 
This command is generate from this file : https://github.com/puppetlabs/puppetlabs-java/blob/master/manifests/oracle.pp.
As it, --proxy is consider as a cookie argument.

bastelfreak:
Hi @kapouik, did you see the comment from @ekohl? I also assume that the rootcause is broken escaping somewhere.

PR #351, opened 2018-08-22
Fallback to PowerShell for zip files on Windows

GeoffWilliams:
Since most users on Windows are going to be using zip files and might not be able to easily install 7zip, how about this approach for using PowerShell/.net built-in support to for zip files to do the extract when there is no 7zip rather then failing? 

~~~
For Windows since 2012, it is possible to extract zip files (only) using
`System.IO.Compression.FileSystem`. For systems where 7zip is not
available, use an inline powershell script to perform the extact. 7zip
will still be used in preference, if available.


bastelfreak:
Thanks for the PR @GeoffWilliams !

PR #329, opened 2018-04-28
rewrite functions using new api

applewiskey:
Rewrite functions using new function API
    Fixes #321 
    Fixes #320 
    Fixes #225 (?)

PR #390, opened 2019-12-06
Convert `assemble_nexus_url` to modern API

alexjfisher:


alexjfisher:
This is the easy bit of https://github.com/voxpupuli/puppet-archive/pull/329
Easy in so much as this function doesn't do anything unsafe like requiring `puppet_x/bodeco/util` (which isn't an environment safe thing to do).

alexjfisher:
> Due to the rename, is this a breaking change?

Depends on whether you think the original function was ever considered public.  I think probably not.

ekohl:
Probably not

lib/puppet/functions/archive/assemble_nexus_url.rb
```diff
@@ -0,0 +1,26 @@
...
+  # @param nexus_url
+  #   The base nexus URL
+  # @param params
+  #   The query parameters as a hash
+  #
+  # @return [Stdlib::HTTPUrl]
+  #   The assembled URL
+  dispatch :default_impl do
+    param 'Stdlib::HTTPUrl', :nexus_url
+    param 'Hash', :params
```
alexjfisher: I should probably put the return type in here too.

PR #392, opened 2019-12-06
Convert `go_md5` function to modern API

alexjfisher:
I'm not actually sure if anybody uses this or whether this works with
recent GoCD servers.  It's pretty easy to port to the new puppet
function API though.


lib/puppet/functions/archive/go_md5.rb
```diff
@@ -0,0 +1,36 @@
+require_relative '../../../puppet_x/bodeco/util.rb'
+
+# @summary
+#   Retrieves and returns specific file's md5 from GoCD server md5 checksum file
+# @api private
+# @see http://www.thoughtworks.com/products/docs/go/12.4/help/Artifacts_API.html
```
alexjfisher: I took this 'broken' link directly from the old function.  I was going to update with eg. https://api.gocd.org/current/#get-artifact-file, but that says `Available since v14.3.0`.  This is why I suspect this function (and `archive::go`) might be outdated and of limited value until a GoCD user comes along and fixes them up.

spec/defines/go_spec.rb
```diff
@@ -3,10 +3,9 @@
 describe 'archive::go' do
   let(:facts) { { os: { family: 'RedHat' }, puppetversion: '4.4.0' } }
 
-  before do
-    MockFunction.new('go_md5') do |f|
-      f.stub.returns('0d4f4b4b039c10917cfc49f6f6be71e4')
-    end
+  # Mock Puppet V4 API ruby function with a puppet language function equivalent
+  let(:pre_condition) do
+    'function archive::go_md5(String $username, String $password, String $file, Stdlib::HTTPUrl $url) { return \'0d4f4b4b039c10917cfc49f6f6be71e4\' }'
```
ekohl: Maybe use heredoc
```suggestion
    <<-PUPPET
    function archive::go_md5(String $username, String $password, String $file, Stdlib::HTTPUrl $url) {
      return '0d4f4b4b039c10917cfc49f6f6be71e4'
    }
    PUPPET
```

alexjfisher: Done.

spec/functions/archive_go_md5_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  it { is_expected.not_to eq(nil) }
+
+  it 'retreives file md5' do
+    url = 'https://gocd.lan/path/file.md5'
+    uri = URI(url)
+    PuppetX::Bodeco::Util.stubs(:content).with(uri, username: 'user', password: 'pass').returns(example_md5)
+    is_expected.to run.with_params('user', 'pass', 'filea', url).and_return('283158c7da8c0ada74502794fa8745eb')
+  end
```
ghoneycutt: Consider adding a test that demonstrates the error state and hits the `raise` in the function.

ghoneycutt:
Suggest adding an acceptance test that pulls a static file, such as from VP's website, and shows that the md5 is as expected.

spec/functions/archive_go_md5_spec.rb
alexjfisher: I was aiming to limit my changes to just converting the function, but actually, I think you're right that the error handling code is broken.  Fixing that up should be simple enough.

alexjfisher:
> Suggest adding an acceptance test that pulls a static file, such as from VP's website, and shows that the md5 is as expected.

At the moment this module hasn't got any acceptance tests.  This function probably isn't where I'd start.  PRs welcome of course! :)

spec/functions/archive_go_md5_spec.rb
alexjfisher: Fixes in https://github.com/voxpupuli/puppet-archive/pull/392/commits/9c5d6323d894acdf24d8a93bd1f63284fb966ad7

PR #399, opened 2020-01-26
Use curl netrc file instead of `--user`

### LINKED ISSUE #397: Feature request: make password sensitive and hide on fail
Would it be possible to hide the password when the archive resource fails?

Ex. change from 'absent' to 'present' failed: Execution of '/usr/bin/curl https://example.org/test.zip -o /tmp/test.zip_20200122-15071-xnnt9z -fsSLg --max-redirs 5 --user test:**test**' returned 22: curl: (22) The requested URL returned error: 404 Not Found

Something like this:
change from 'absent' to 'present' failed: Execution of '/usr/bin/curl https://example.org/test.zip -o /tmp/test.zip_20200122-15071-xnnt9z -fsSLg --max-redirs 5 --user test:**Sensitive [value redacted]**' returned 22: curl: (22) The requested URL returned error: 404 Not Found

alexjfisher:
This stops credentials appearing in the process list and failed
downloads won't result in the credentials being logged in puppet reports
and agent log output.

Fixes #397

### DIFF
--- a/lib/puppet/provider/archive/curl.rb
+++ b/lib/puppet/provider/archive/curl.rb
@@ -1,16 +1,35 @@
+require 'uri'
+require 'tempfile'
+
 Puppet::Type.type(:archive).provide(:curl, parent: :ruby) do
   commands curl: 'curl'
   defaultfor feature: :posix
 
   def curl_params(params)
-    account = [resource[:username], resource[:password]].compact.join(':') if resource[:username]
-    params += optional_switch(account, ['--user', '%s'])
+    if resource[:username]
+      create_netrcfile
+      params += ['--netrc-file', @netrc_file.path]
+    end
     params += optional_switch(resource[:proxy_server], ['--proxy', '%s'])
     params += ['--insecure'] if resource[:allow_insecure]
     params += resource[:download_options] if resource[:download_options]
     params += optional_switch(resource[:cookie], ['--cookie', '%s'])
 
     params
+  end
+
+  def create_netrcfile
+    @netrc_file = Tempfile.new('.puppet_archive_curl')
+    machine = URI.parse(resource[:source]).host
+    @netrc_file.write("machine #{machine}\nlogin #{resource[:username]}\npassword #{resource[:password]}\n")
+    @netrc_file.close
+  end
+
+  def delete_netrcfile
+    return if @netrc_file.nil?
+
+    @netrc_file.unlink
+    @netrc_file = nil
   end
 
   def download(filepath)
@@ -25,7 +44,11 @@ Puppet::Type.type(:archive).provide(:curl, parent: :ruby) do
       ]
     )
 
-    curl(params)
+    begin
+      curl(params)
+    ensure
+      delete_netrcfile
+    end
   end
 
   def remote_checksum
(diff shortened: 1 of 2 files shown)


alexjfisher:
I intend to fixup/rewrite the tests after getting some initial feedback.

alexjfisher:
Should probably implement a similar fix for the wget provider.  Annoyingly, wget can also use netrc files, but it appears you can't actually specify a path to such a file.  It only looks in ~/.netrc (which we can't overwrite!)

alexjfisher:
You can use a WGETRC environment variable.
https://www.gnu.org/software/wget/manual/html_node/Wgetrc-Location.html

Since the implementation wouldn't have too much in common, and since you have to go out of your way to select wget as the provider, I'll probably just leave this for a separate PR.

igalic:
and we can't use Sensitive type in a puppet type?

alexjfisher:
Type *properties* automatically work with Sensitive types.  When puppet syncs the property it just knows to redact the change.  But `password` is a *parameter*.  (You can also define a property to be Sensitive always even if the user provides a normal string.  See https://github.com/puppetlabs/puppet/commit/f62a5920f0a7c60136ccee97f14d46554364f125)

With parameters, you can pass a Sensitive value if you like and it will work.  But you'll get a warning like:

> Unable to mark 'password' as sensitive: password is a parameter and not a property, and cannot be automatically redacted.

I suppose you could suppress the warning by converting the parameter into a property that is fudged to always be in sync (like in https://github.com/puppetlabs/puppetlabs-stdlib/pull/786)

But since we don't (in this fix of this provider), pass the password to the command being executed, we don't have to do something crazy like in https://github.com/puppetlabs/puppetlabs-vcsrepo/pull/416/files

alexjfisher:
By switching to using a netrc file, I think we avoid all the hassle of having to conditionally redact command output by calling Puppet::Util::Execution directly instead of using the autogenerated helper.  If we couldn't have used netrc, then I don't think I'd even check to see if password was Sensitive, but *always* consider it so.
I'm also guessing if you `Puppet::Util::Execution.execute("curl $PARAMS", sensitive: true)` you end up redacting the command error output completely.  This would be very helpful to users.

alexjfisher:
I'm unclear as to why in https://github.com/puppetlabs/puppetlabs-vcsrepo/pull/416/files#diff-574463866055cc3f49ee013638f6652aR81 they have to call `unwrap`.  If I set archive's password parameter to a Sensitive value, it just works.  No need to explicitly unwrap anything.  It gets written out to the file just fine.

lib/puppet/provider/archive/curl.rb
```diff
@@ -25,7 +44,12 @@ def download(filepath)
       ]
     )
 
-    curl(params)
+    begin
+      curl(params)
+    rescue StandardError => e
```
ekohl: You can also use `ensure` which is more suited for this particular case.

ekohl:
I like this approach. One reason to do this regardless of the use of `Sensitive` is that it doesn't show the password in the process list (`ps aux`).

lib/puppet/provider/archive/curl.rb
alexjfisher: and more to the point, I've completely forgotten to call delete_netrcfile when the command didn't fail!!  Doh!

alexjfisher: Fixed.

ekohl:
Thanks!

PR #406, opened 2020-04-02
No need for Augeas PPA on bionic

raphink:
Ubuntu Bionic has Augeas 1.10 already available, so there is no need to use the Augeas PPA (which provides Augeas 1.16)


alexjfisher:
Duplicate of #405

PR #411, opened 2020-07-25
modulesync 3.0.0 & puppet-lint updates

bastelfreak:
modulesync 3.0.0


alexjfisher:
Replaced with https://github.com/voxpupuli/puppet-archive/pull/413

PR #412, opened 2020-08-11
Clean up temporary files when checksums don't match

benridley:
Noticed that the archive module was leaving behind quite a few files in /tmp when there were checksum mismatches.

This ensures that the temporary files will be cleaned up before throwing an error.

This pull request removes the temporary files created by the archive module if the checksum fails to match. Previously, the module would throw an error and leave the temporary file in /tmp.

### DIFF
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -191,6 +191,7 @@ Puppet::Type.type(:archive).provide(:ruby) do
       archive = PuppetX::Bodeco::Archive.new(temppath)
       actual_checksum = archive.checksum(resource[:checksum_type])
       if actual_checksum != checksum
+        destroy(temppath)
         raise(Puppet::Error, "Download file checksum mismatch (expected: #{checksum} actual: #{actual_checksum})")
       end
     end


ghoneycutt:
@benridley Good idea! Could you please add a test so this can be merged. Acceptance tests are preferred. You can cause the situation where a tmp file would normally be left then check the existence of the file to show that it is not there.

benridley:
Hey @ghoneycutt, happy to add an acceptance test to get this merged. 

I'm not really too familiar with Ruby testing. Is RSpec the right path to go down here or is that strictly unit testing? Cheers!

ghoneycutt:
This module does not have acceptance tests setup and the errors are totally unrelated to this so merging.

igalic:
> This module does not have acceptance tests setup and the errors are totally unrelated to this so merging.

would this module benefit from acceptance tests?

ghoneycutt:
Ha, it sure would! Are you volunteering :)

lib/puppet/provider/archive/ruby.rb
```diff
@@ -191,6 +191,7 @@ def transfer_download(archive_filepath)
       archive = PuppetX::Bodeco::Archive.new(temppath)
       actual_checksum = archive.checksum(resource[:checksum_type])
       if actual_checksum != checksum
+        destroy(temppath)
```
alexjfisher, line 194: `destroy` doesn’t take any arguments...
Fixed in #443

PR #415, opened 2020-08-18
Add support param $temp_dir in archive::nexus

alexcit:

### DIFF
--- a/manifests/nexus.pp
+++ b/manifests/nexus.pp
@@ -44,6 +44,8 @@ define archive::nexus (
   Optional[String]  $proxy_server    = undef,
   Optional[String]  $proxy_type      = undef,
   Optional[Boolean] $allow_insecure  = undef,
+  Optional[String]  $proxy_type      = undef,
+  Optional[String]  $temp_dir        = undef,
 ) {
   include archive::params
 
@@ -107,6 +109,7 @@ define archive::nexus (
     proxy_server    => $proxy_server,
     proxy_type      => $proxy_type,
     allow_insecure  => $allow_insecure,
+    temp_dir        => $temp_dir,
   }
 
   $file_owner = pick($owner, $archive::params::owner)


manifests/nexus.pp
```diff
@@ -44,6 +44,8 @@
   Optional[String]  $proxy_server    = undef,
   Optional[String]  $proxy_type      = undef,
   Optional[Boolean] $allow_insecure  = undef,
+  Optional[String]  $proxy_type      = undef,
```
igalic, line 47: you're repeating a param that's there, two lines above

igalic (changes requested):
this is why tests are failing, but if you're adding a new parameter, i'd like to see a new test, too

manifests/nexus.pp
alexcit: Sorry, my fault. I should be more careful. I have removed the duplicate parameter

igalic (changes requested):
please add the new parameter to the tests here
https://github.com/voxpupuli/puppet-archive/blob/master/spec/defines/nexus_spec.rb

alexcit:
> please add the new parameter to the tests here
> https://github.com/voxpupuli/puppet-archive/blob/master/spec/defines/nexus_spec.rb
Is it Ok if I'll add in ['nexus archive with overwritten parameters'](https://github.com/voxpupuli/puppet-archive/blob/master/spec/defines/nexus_spec.rb#L33) part and won't create a new code block?

igalic:
yes

spec/defines/nexus_spec.rb
```diff
@@ -44,7 +44,8 @@
         extract: true,
         extract_path: '/opt',
         creates: '/opt/artifact/WEB-INF',
-        cleanup: true
+        cleanup: true,
+        temp_dir : '/tmp'
```
igalic, line 48: this is not how [symbolised hashes in Ruby](https://ruby-doc.org/core-2.7.1/Hash.html) work
there must not be a space between the word, and the `:`

igalic (changes requested):
🙋🏻‍♀️

spec/defines/nexus_spec.rb
alexcit: Thx, I have fixed it

igalic (approved):
👍

manifests/nexus.pp
```diff
@@ -44,6 +44,7 @@
   Optional[String]  $proxy_server    = undef,
   Optional[String]  $proxy_type      = undef,
   Optional[Boolean] $allow_insecure  = undef,
+  Optional[String]  $temp_dir        = undef,
```
alexjfisher, line 47: I'd have probably gone for `Optional[Stdlib::Absolutepath]`

alexcit: As you wish :)

PR #421, opened 2020-08-24
Add initial support for gsutil and pulling from Google Storage buckets

j0sh3rs:
This PR adds support in the archive module for pulling assets from Google Storage buckets by way of gsutil

No Issues were filed prior to this PR.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,12 +23,15 @@
 #   Alternative package source for 7zip.  This parameter only applies to Windows.
 # @param aws_cli_install
 #   Installs the AWS CLI command needed for downloading from S3 buckets.  This parameter is currently not implemented on Windows.
+# @param gsutil_install
+#   Installs the GSUtil CLI command needed for downloading from GS buckets.  This parameter is currently not implemented on Windows.
 #
 class archive (
   Optional[String[1]]                       $seven_zip_name     = $archive::params::seven_zip_name,
   Optional[Enum['chocolatey','windows','']] $seven_zip_provider = $archive::params::seven_zip_provider,
   Optional[String[1]]                       $seven_zip_source   = undef,
   Boolean                                   $aws_cli_install    = false,
+  Boolean                                   $gsutil_install     = false,
 ) inherits archive::params {
   if $facts['os']['family'] == 'Windows' and !($seven_zip_provider in ['', undef]) {
     package { '7zip':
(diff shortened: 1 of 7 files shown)


bastelfreak:
@j0sh3rs thanks for the PR. Can you add some unit tests for this change?

j0sh3rs:
@bastelfreak -- Added some tests, let me know if they're up to par! I basically copied the ruby spec and modified it for gsutil.

manifests/init.pp
```diff
@@ -66,4 +69,32 @@
...
+    # TODO: Windows support.
+    if $facts['os']['family'] != 'Windows' {
+      # Using bundled install option:
+      # https://cloud.google.com/storage/docs/quickstart-gsutil
+
+      file { '/opt/gsutil-bundle':
+        ensure => 'directory',
+      }
+
+      archive { 'gsutil.zip':
```
bastelfreak, line 83: I was really hoping to find packages for linux distros that we could manage instead, but I failed :(

bastelfreak (approved):
looks good to me, but I'm not a types/provider expert. @alexjfisher or @ekohl would you like to review as well?

manifests/init.pp
igalic: it's Google.

manifests/init.pp
```diff
@@ -66,4 +69,32 @@
       }
     }
   }
+
+  if $gsutil_install {
+    # TODO: Windows support.
```
igalic, line 74: https://cloud.google.com/storage/docs/gsutil_install?hl=en#windows
installation on windows uses the same archive, but different paths
and python installation on windows isn't guaranteed, so that's another issue, and so might the execution be

however, given that this thing requires python 2.7, which is EOL, a safe execution on any platform is not guaranteed

igalic (approved):
the provider and the tests look good
I'm just not sure about gsutil itself

manifests/init.pp
mgaidia: https://github.com/GoogleCloudPlatform/gsutil/issues/29#issuecomment-525063367

igalic: would be cool if they updated their documentation

j0sh3rs: Hi @igalic -- totally agreed that the doc on Google's side could use some work (this is seemingly systemic for them). Are we considering this a blocker for acceptance of this PR, or is there something additional I can provide to help shepherd this through?

Thanks very much for the review.

igalic: it was a remark on my approving of this pr

so i think with enough eyes on this now, we can hit the merge button and see what happens *after* that

Issue #400, opened 2020-02-06
Download fails when host cert is signed by private CA


BSipos-RKF:
## Affected Puppet, Ruby, OS and module versions/distributions
- Puppet: 5.3.10
- Ruby: 2.0.0.648
- Distribution: CentOS 7
- Module version: 4.4.0

## How to reproduce (e.g Puppet code you use)
Use an archive resource with a "source" URL of a host with HTTPS certificate signed through a private CA chain.
The operating system (Windows 10) has the private root and and intermediate CAs trusted at the operating system level.

## What are you seeing
The pupet agent gives an error about not being able to validate the host certificate.

## What behaviour did you expect instead
The puppet agent should use the OS-level CA trust to validate the host cert.

BSipos-RKF:
A workaround is to set "allow_insecure => true" but this defeats the authentication of TLS.

Mystakill:
"allow_insecure => true" also doesn't work when the endpoint being connected to is enforcing SSL/TLS for requestor authentication & enforcing permissions.  That makes this a non-starter for us, as all of the hosts in our enterprise do this.  We need a way to specify the certificate & key .pem files, and either (preferably) a capath containing all of the certificate authority certs, or a cacert file contain all of the authority certs combined.

kenyon:
See also: #188

Issue #188, opened 2016-07-15
Add new `ca_certificate` parameter


alexjfisher:
As an alternative to setting `allow_insecure` to `true`, some users might prefer to specify their own cacerts file.

For `wget` the option is `--ca-certificate`
For `curl` the option is `--with-ca-bundle`
For `ruby` `http.ca_file`?

igalic:
imo, you should probably ~*~just~*~ manage your ca stores "correctly" and have your CAs in (***all***) ca stores

alexjfisher:
I've taken a closer look.  The ruby provider currently sets the `SSL_CERT_FILE` environment variable to the cacerts.pem file the module ships, but only for Windows and only if that environment variable was not previously set.  Being able to use a user provided file would be an improvement here.  I'm not sure what the history is and how/why Windows differs from Linux.

For curl/wget, I guess ideally, users would always have all their ca stores setup correctly, but providing this functionality might still be useful to some?  If I find a bit of spare time, implementing this feature for these providers should be fairly trivial.  Perhaps somebody later could then fix up the ruby provider and probably implement `allow_insecure` at the same time.

nanliu:
TLDR; this was to work around a cert issue in some version of Windows Ruby. This was never intended as a feature.

Specifically this issue: https://gist.github.com/luislavena/f064211759ee0f806c88

Basically, a version of Windows ruby shipped a cacert that wouldn't even connect to rubygems.org. When this module had gem dependencies, this work around was required to bootstrap itself. At this point, we no longer have gem dependencies and hopefully the Ruby packages on windows does not suffer from that bug anymore, so we should be able to remove this. I agree with @igalic that the cacerts should be handled by the system and not embedded in a puppet module.

LTangaF:
I actually recently encountered this bug and have a PR that I'm going to be creating here on that.  The environment variable SSL_CERT_FILE on Windows seems to be being ignored by Net::HTTP and I've coded a workaround that sets it specifically on the http instance before 'start'ing it.  Puppet's shipped environment.bat seems to set that environment variable to a non-existing file in C:\Windows\System32\ssl by default and we've had to make sure that file was created, too.

PR #422, opened 2020-10-01
Support for cacert_file param on archive resource

benningm:

This adds a cacert_file parameter to the archive type and
support for this option in curl, wget and ruby provider.

The option could be used to specify a custom CA bundle for
certificate verification in the TLS handshake.

related #400 
implements #188 

#### Additional notes

I'm unsure about the part in lib/puppet_x/bodeco/util.rb with ENV['SSL_CERT_FILE'].
Where is it used? Is it still required?

### DIFF
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -237,7 +237,7 @@ Puppet::Type.newtype(:archive) do
   newparam(:cacert_file) do
     desc 'path to a custom CA certificate bundle file'
     validate do |value|
-      unless !value.nil? && Puppet::Util.absolute_path?(value)
+      if !value.nil? && !Puppet::Util.absolute_path?(value)
         raise ArgumentError, "cacert_file must be absolute: #{value}"
       end
     end
--- a/lib/puppet/type/archive.rb
+++ b/lib/puppet/type/archive.rb
@@ -234,6 +234,15 @@ Puppet::Type.newtype(:archive) do
     defaultto :false
   end
 
+  newparam(:cacert_file) do
+    desc 'path to a custom CA certificate bundle file'
+    validate do |value|
+      unless !value.nil? && Puppet::Util.absolute_path?(value)
+        raise ArgumentError, "cacert_file must be absolute: #{value}"
+      end
+    end
+  end
+
   newparam(:download_options) do
     desc 'provider download options (affects curl, wget, gs, and only s3 downloads for ruby provider)'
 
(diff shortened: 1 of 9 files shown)


bastelfreak (approved):
looks good to me, but I'm far away from being an expert for archive. Somebody else should review this as well.

alexjfisher (approved):
LGTM

lib/puppet/type/archive.rb
```diff
@@ -234,6 +234,15 @@ def should_to_s(value)
     defaultto :false
   end
 
+  newparam(:cacert_file) do
+    desc 'path to a custom CA certificate bundle file'
+    validate do |value|
+      unless !value.nil? && Puppet::Util.absolute_path?(value)
```
ekohl, line 240: Is this correct? I feel like it should be:
```suggestion
      if !value.nil? && !Puppet::Util.absolute_path?(value)
```

However, I'm having a hard time wrapping my brain on it. It would be great if you could just say `type 'Optional[Stdlib::AbsolutePath]'` in the type DSL.

lib/puppet_x/bodeco/util.rb
```diff
@@ -80,7 +80,13 @@ def initialize(_url, options)
           @proxy_port = uri.port
         end
 
-        ENV['SSL_CERT_FILE'] = File.expand_path(File.join(__FILE__, '..', 'cacert.pem')) if Facter.value(:osfamily) == 'windows' && !ENV.key?('SSL_CERT_FILE')
+        @cacert_file = if options.key?[:cacert_file]
+                         options[:cacert_file]
+                       elsif ENV.key?('SSL_CERT_FILE')
+                         ENV['SSL_CERT_FILE']
+                       elsif Facter.value(:osfamily) == 'windows'
+                         File.expand_path(File.join(__FILE__, '..', 'cacert.pem'))
```
ekohl, line 88: This can be simplified:
```suggestion
                         File.join(__dir__, 'cacert.pem')
```

However, is there ever a `cacert.pem` in the directory of this file?

benningm: Sure, but i did not change that part. Just moved it to a different location in the logic.

benningm (commented):
Regarding the type validation i copied that over from the other path options. Maybe you change all the options to use puppet type validations.

ekohl (approved):

ekohl:
@alexjfisher any last comments?

bastelfreak:
Hi @benningm . I merged this and another PR that introduced tests. code from #422 failed with the new tests so I had to revert it for now. Please check https://github.com/voxpupuli/puppet-archive/pull/440 for details. can you submit the PR again and take a look at the failing tests?

benningm:
If you can confirm that it will not take a half year to review it and then will be reverted again.

PR #423, opened 2020-10-21
Add `archives` parameter to make use with an ENC or hiera easier

jcpunk:

This PR adds a parameter to make generating archive requests from an ENC easier.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -15,6 +15,14 @@
 #     aws_cli_install => true,
 #   }
 #
+# @example Deploy a specific archive
+#   class { 'archive':
+#     archives => { '/tmp/jta-1.1.jar' => {
+#                     'ensure' => 'present',
+#                     'source'  => 'http://central.maven.org/maven2/javax/transaction/jta/1.1/jta-1.1.jar',
+#                     }, }
+#   }
+#
 # @param seven_zip_name
 #   7zip package name.  This parameter only applies to Windows.
 # @param seven_zip_provider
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -32,6 +32,7 @@
   Optional[String[1]]                       $seven_zip_source   = undef,
   Boolean                                   $aws_cli_install    = false,
   Boolean                                   $gsutil_install     = false,
```
alexjfisher, line 34: Best to default to an empty hash instead of `undef`

manifests/init.pp
```diff
@@ -97,4 +98,10 @@
       }
     }
   }
+
+  $archives.each |$archive_name, $archive_settings| {
+    archive { $archive_name
```
alexjfisher, line 103: missing a `:` at the end of this line.

alexjfisher (changes requested):
Thanks for the PR. I'm not against adding this functionality, but it currently has a syntax error.

manifests/init.pp
```diff
@@ -32,6 +32,7 @@
   Optional[String[1]]                       $seven_zip_source   = undef,
   Boolean                                   $aws_cli_install    = false,
   Boolean                                   $gsutil_install     = false,
+  Optional[Hash]                            $archives           = {},
```
alexjfisher, line 35: The data type can be just `Hash` now.

jcpunk: I'll fix an push new.

manifests/init.pp
```diff
@@ -32,6 +32,7 @@
   Optional[String[1]]                       $seven_zip_source   = undef,
   Boolean                                   $aws_cli_install    = false,
   Boolean                                   $gsutil_install     = false,
+  Hash                                      $archives           = {},
```
alexjfisher, line 35: Could you document the parameter using the puppet-strings style as above https://github.com/voxpupuli/puppet-archive/pull/423/files#diff-c6c3a824d185ff01e063011517a1b13a604dc3d7b63d1c67a30451c10df02bcbL26?

I dunno, something like...

```puppet
# @param archives
#   A hash of archive resources that this module should create
```

Bonus points for adding a `@example` to the examples section.

jcpunk: Done

alexjfisher (approved):

kenyon:
Isn't this for hiera, not an ENC?

alexjfisher:
> Isn't this for hiera, not an ENC?

either.

Issue #429, opened 2021-01-16
Cannot pull password-protected HTTP URLs after merge of netrc


frozenfoxx:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.19.1
- Ruby: 2.5.8p224
- Distribution: Ubuntu 18.04
- Module version: 4.6.0

## How to reproduce (e.g Puppet code you use)

```
  # Pull archive for configuration and extract if newer
  archive { "${facts['networking']['hostname']}.tar.gz":
    ensure          => present,
    path            => "${datasourcedir}/${facts['networking']['hostname']}.tar.gz",
    source          => $archiveurl,
    username        => $archiveuser,
    password        => $archivepassword,
    extract         => true,
    extract_path    => "${datasourcedir}/tool",
    extract_command => 'tar xfz %s --strip-components=1',
    creates         => "${datasourcedir}/file_that_exists_after_extract",
    notify          => Docker::Run['tool']
  }
```

## What are you seeing

Fails to download due to a 401 Unauthorized.

## What behaviour did you expect instead

Download of the archive and extraction.

## Output log

```
Error: Execution of '/usr/bin/curl https://company.jfrog.io/artifactory/project/team/tool/archives/fqdn.tar.gz -o /tmp/fqdn.tar.gz_20210116-9592-1yt2mha -fsSLg --max-redirs 5 --netrc-file /tmp/.pupp
et_archive_curl20210116-9592-4uamxq' returned 22: curl: (22) The requested URL returned error: 401
Error: /Stage[main]/Tool::Install/Archive[fqdn.tar.gz]/ensure: change from 'absent' to 'present' failed: Execution of '/usr/bin/curl https://company.jfrog.io/artifactory/project/team/tool/archives/fqdn.tar.gz -o /tmp/fqdn.tar.gz_20210116-9592-1yt2mha -fsSLg --max-redirs 5 --netrc-file /tmp/.puppet_archive_curl20210116-9592-4uamxq' returned 22: curl: (22) The requested URL returned error: 401
```

## Any additional information you'd like to impart

This works perfect with 4.5.0. Looks like it's having some sort of an issue with the `netrc` file, but since it destroys the file after it fails I can't see what's going in it. I've had to rollback. Seems to be related to this change: https://github.com/voxpupuli/puppet-archive/pull/399.

alexjfisher:
[...]

PR #430, opened 2021-01-22
Fix downloading when passwords contain spaces

alexjfisher:
This also adds our first acceptance tests to this module.

### DIFF
--- a/lib/puppet/provider/archive/curl.rb
+++ b/lib/puppet/provider/archive/curl.rb
@@ -7,8 +7,14 @@ Puppet::Type.type(:archive).provide(:curl, parent: :ruby) do
 
   def curl_params(params)
     if resource[:username]
-      create_netrcfile
-      params += ['--netrc-file', @netrc_file.path]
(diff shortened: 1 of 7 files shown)


lib/puppet/provider/archive/curl.rb
```diff
@@ -7,8 +7,14 @@
 
   def curl_params(params)
     if resource[:username]
-      create_netrcfile
-      params += ['--netrc-file', @netrc_file.path]
+      if resource[:username] =~ %r{\s} || resource[:password] =~ %r{\s}
+        Puppet.warning('Username or password contains a space.  Unable to use netrc file to hide credentials')
+        account = [resource[:username], resource[:password]].compact.join(':')
```
alexjfisher, line 12: This is the implementation from before https://github.com/voxpupuli/puppet-archive/pull/399

alexjfisher:
There's a small chance that this is the fix for the issue reported in https://github.com/voxpupuli/puppet-archive/issues/429

bastelfreak (approved):

PR #431, opened 2021-01-22
Drop support for CentOS/RHEL 6 and variants

alexjfisher:
CentOS 6 is now EOL.

There's little reason to think this module won't continue to work for
some time, but this change is also a prerequisite for enabling
acceptance tests, (where it's no longer possible to test on CentOS 6)

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -3,7 +3,6 @@
     {
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
-        "6",
         "7",
         "8"
       ]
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -3,33 +3,28 @@
...
         "7",
         "8"
       ]
     },
     {
       "operatingsystem": "Scientific",
       "operatingsystemrelease": [
-        "6",
-        "7",
-        "8"
```
alexjfisher, line 32: Scientific Linux 8 is not a thing, (at the moment anyway!!)

PR #433, opened 2021-02-09
Pass over credentials in archive::artifactory

jramosf:
This change allows passing optional credentials to the `archive` resource from `archive::artifactory`.

### DIFF
--- a/manifests/artifactory.pp
+++ b/manifests/artifactory.pp
@@ -50,6 +50,8 @@ define archive::artifactory (
   Optional[String]               $extract_path = undef,
   Optional[String]               $creates      = undef,
   Optional[Boolean]              $cleanup      = undef,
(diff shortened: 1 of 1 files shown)


jramosf:
Hello @kenyon ,

You added the `needs-tests` label but I see #430 is already [testing](https://github.com/voxpupuli/puppet-archive/blob/0eaeeaa1bcceeae5f3e5bd4977d5a0f2bca4a791/spec/acceptance/authentication_spec.rb) the password functionality of `archive`, and this is just passing over the parameters to that same function.

Cheers,

kenyon (approved):

Rwwn:
This doesn't work for me because the artifactory_checksum function also requires separate username/password authentication, which isn't covered in this PR. Providing the username and password still results in a forbidden error.

PR #434, opened 2021-03-02
Produce a better error for the puppet downloader when file not found

hajee:
Newer versions of Puppet raise the error Puppet::HTTP::RouteError when the file is not found. This resulted not only in a misleading error but also an error without the original url.

This commit changes this. It catches the error and produces an error with the url  in it. Unfortunately, the original misleading error is still produces, so we try to elaborate a bit in the error we produce.


hajee:
Is there an ETA when there is time to look at this PR?

kenyon:
> Is there an ETA when there is time to look at this PR?

No, just when one of us @voxpupuli volunteers gets around to it. :smile:

The change seems fine to me, but I don't use this module.

bastelfreak:
thanks for the PR!

PR #428, opened 2020-12-01
Windows: find 7zip binary

joerg16:
#### Pull Request (PR) description
Search for 7z.exe Binary if not installed in the standard Paths

PR #440, opened 2021-04-16
Revert "Support for cacert_file param on archive resource"

bastelfreak:
Reverts voxpupuli/puppet-archive#422

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,7 +464,6 @@ archive { '/tmp/staging/master.zip':
 * `source`: archive file source, supports http|https|ftp|file|s3|gs uri.
 * `username`: username to download source file.
 * `password`: password to download source file.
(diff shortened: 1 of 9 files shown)


bastelfreak:
As you can see in https://github.com/voxpupuli/puppet-archive/pull/439, master fails since the merge of https://github.com/voxpupuli/puppet-archive/pull/430. The thing is that 430 introduced new tests. So I guess one of the three PRs broke it but that wasn't detected because tests were missing:
* https://github.com/voxpupuli/puppet-archive/pull/434
* https://github.com/voxpupuli/puppet-archive/pull/422
* https://github.com/voxpupuli/puppet-archive/pull/428

bastelfreak:
tests are green again, so I guess it's safe to say that this code broke it

PR #444, opened 2021-05-28
Update CA certificate bundle to 2021-05-25

l-avila:


kenyon:
Source: https://curl.se/docs/caextract.html

PR #449, opened 2021-07-30
Set optional param to undef to fix failling test

### LINKED ISSUE #450: Fix archive::download Param Default Value

- Puppet: 6 & 7
- Ruby: 2.5 & 2.7
- Distribution: N/A
- Module version: 5.0.0

Run `bundle exec rake test` using the 5.0.0 tag version of the module.

## What are you seeing
```
❯ bundle exec rake test
manifests/download.pp:38:optional_default:WARNING:Optional parameter defaults to something other than undef
```

Passing tests.

## Any additional information you'd like to impart
Optional parameters should default to undef, so I think fixing this behavior will result in a backwards-incompatible change, since existing users not explicitly setting the `digest_type` of `archive::download`, will need to do so after this change.

yachub:
This fixes a test failure detected by the recently added GitHub Actions workflow.

Fixes #450

### DIFF
--- a/manifests/download.pp
+++ b/manifests/download.pp
@@ -35,7 +35,7 @@ define archive::download (
   Boolean                       $checksum         = true,
   Optional[String]              $digest_url       = undef,
   Optional[String]              $digest_string    = undef,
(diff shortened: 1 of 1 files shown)


genebean:
Holding merge while an associated issue is filed and the commit message is updated.

genebean (approved):

PR #467, opened 2021-12-10
Use ssl_trust_store from puppet.conf on Windows

southalc:
The Windows Puppet agent fails to download archive resources due to SSL validation failure when the "source" is using a certificate issued by a private CA.  The failure occurs even when the Puppet agent is configured with a custom "ssl_trust_store" that contains the CA chain.

This patch changes the Windows download behavior by defining the following order for the SSL trust store:
1. The "ssl_trust_store" setting from the "agent" section of "puppet.conf"
2. The "SSL_CERT_FILE" value from the runtime environment
3. The module-bundled ".pem" file as a last resort.

Fixes issue reported at: https://tickets.puppetlabs.com/browse/PUP-11349

### DIFF
--- a/lib/puppet_x/bodeco/util.rb
+++ b/lib/puppet_x/bodeco/util.rb
@@ -80,7 +80,22 @@ module PuppetX
           @proxy_port = uri.port
         end
 
-        ENV['SSL_CERT_FILE'] = File.expand_path(File.join(__FILE__, '..', 'cacert.pem')) if Facter.value(:osfamily) == 'windows' && !ENV.key?('SSL_CERT_FILE')
+        @osfamily = Facter.value(:osfamily)
+        if @osfamily == 'windows'
+          # Get the 'ssl_trust_store' setting from the puppet agent
+          Puppet.settings.preferred_run_mode = :agent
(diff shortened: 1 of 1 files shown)


southalc:
Checking in on this pull request.  I noticed the "needs-tests" label was added, but it doesn't look like the current implementation has any unit tests for SSL connections.  This change maintains the current behavior and only enables an override of the default SSL trust file used by the Windows agent.  What are we looking for in tests?

albatrossflavour:
@ghoneycutt - I grok the request for tests, but... we also need to move stuff forward.  How about we approve this and we'll work out how to get some time from the Puppet team to work on adding tests?

kenyon:
@albatrossflavour just to be clear, @voxpupuli modules like this one aren't maintained by @puppetlabs (@perforce). We are all volunteers.

albatrossflavour:
@kenyon I'm very clear on that, which is why I'm offering some support from my team to help.  If we work together, we can get more done and develop the skills of more practitioners (internal and community)

kenyon:
@albatrossflavour OK, I guess you work for Puppet, which would make more sense, but that wasn't obvious 🙂

phiggins2:
any chance of this progressing in 2025?

PR #470, opened 2022-04-25
Fix errors and improve README

### LINKED ISSUE #469: Example in README doesn't make sense.
The following [code in lines 228-238 of README.md](https://github.com/voxpupuli/puppet-archive/blob/master/README.md?plain=1#L228-L238) has the same [`$install_path`](https://github.com/voxpupuli/puppet-archive/blob/master/README.md?plain=1#L221) path listed for both the `creates` and `require` attributes:
```
archive { $filename:
  path          => "/tmp/${filename}",
  source        => "http://www-eu.apache.org/dist/tomcat/tomcat-9/v9.0.0.M3/bin/apache-tomcat-9.0.0.M3.zip",
  checksum      => 'f2aaf16f5e421b97513c502c03c117fab6569076',
  checksum_type => 'sha1',
  extract       => true,
  extract_path  => '/opt',
  creates       => $install_path,
  cleanup       => 'true',
  require       => File[$install_path],
}
```

If `File[$install_path]` is declared elsewhere and evaluated successfully, the `creates` attribute should prevent the archive from being downloaded, according to [line 492](https://github.com/voxpupuli/puppet-archive/blob/master/README.md?plain=1#L492):
> * `creates`: if file/directory exists, will not download/extract archive.

If `File[$install_path]` is not declared elsewhere, catalog compilation should fail because of the [missing dependency](https://puppet.com/docs/puppet/7/lang_relationships.html#missing-dependencies):

> If one of the resources in a relationship is never declared, compilation fails with one of the following errors:
> * Could not find dependency &lt;OTHER RESOURCE&gt; for &lt;RESOURCE&gt;
> * Could not find resource &lt;OTHER RESOURCE&gt; for relationship on &lt;RESOURCE&gt;

If `File[$install_path]` is declared elsewhere but [fails](https://puppet.com/docs/puppet/7/lang_relationships.html#lang_relationships-failed-dependencies), the archive resource should be skipped:

> If Puppet fails to apply the prior resource in a relationship, it skips the subsequent resource and log the following messages:
> `notice: <RESOURCE>: Dependency <OTHER RESOURCE> has failures: true warning: <RESOURCE>: Skipping because of failed dependencies`
> It then continues to apply any unrelated resources. Any resources that depend on the skipped resource are also skipped. This helps prevent an inconsistent system state, rather than attempting to apply a resource that might have broken prerequisites.

The example as written doesn't appear to make sense and should be removed or corrected.

pillarsdotnet:

- Improve documentation of `creates` attribute.
- Specify Enum attributes as symbols instead of text strings according to current best-practice.

Fixes #469

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ archive { '/tmp/jta-1.1.jar':
   extract_path  => '/tmp',
   source        => 'http://central.maven.org/maven2/javax/transaction/jta/1.1/jta-1.1.jar',
   checksum      => '2ca09f0b36ca7d71b762e14ea2ff09d5eac57558',
-  checksum_type => 'sha1',
+  checksum_type => sha1,
   creates       => '/tmp/javax',
   cleanup       => true,
 }
@@ -202,7 +202,7 @@ archive { $filename:
   path          => "/tmp/${filename}",
   source        => 'http://www-eu.apache.org/dist/tomcat/tomcat-9/v9.0.0.M3/bin/apache-tomcat-9.0.0.M3.zip',
   checksum      => 'f2aaf16f5e421b97513c502c03c117fab6569076',
-  checksum_type => 'sha1',
+  checksum_type => sha1,
   extract       => true,
   extract_path  => '/opt',
   creates       => "${install_path}/bin",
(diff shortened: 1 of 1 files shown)


ghoneycutt:
Thanks @pillarsdotnet

PR #474, opened 2022-09-29
ruby provider: ensure cleanup happens

### LINKED ISSUE #328: Cannot clean up unless 'creates' is specified.
It appears that the archive file will not be cleaned up unless the 'creates' parameter is used AND the file that is specified by that parameter exists after extraction. 

I think it would be useful to be able to cleanup even if this parameter is not specified or does not seem to exist. Is there reasoning behind not doing this?

pillarsdotnet:
Ensures that tempfile used for downloading gets deleted.
Ensures that the `cleanup` param is honored.

Fixes #328

### DIFF
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -84,6 +84,7 @@ Puppet::Type.type(:archive).provide(:ruby) do
   def create
     transfer_download(archive_filepath) unless checksum?
     extract
+  ensure
     cleanup
   end
 
@@ -144,8 +145,7 @@ Puppet::Type.type(:archive).provide(:ruby) do
   end
 
   def cleanup
-    return unless extracted? && resource[:cleanup] == :true
-
+    return unless resource[:cleanup] == :true and resource[:extract] == :true
     Puppet.debug("Cleanup archive #{archive_filepath}")
     destroy
   end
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/archive/ruby.rb
```diff
@@ -80,6 +80,7 @@ def exists?
   def create
     transfer_download(archive_filepath) unless checksum?
     extract
+  ensure
```
bastelfreak, line 83: whats enaure doing here, is that a ruby internal as well? if its a simple method call, shouldnt it be indented more?

pillarsdotnet: Yes, `ensure` is a ruby keyword. It marks a block of code that must run before the method exits, even if an exception is raised.  It gets indented the same way as a `rescue` block.

bastelfreak:
Hi @pillarsdotnet, thanks for the PR. can you please rebase against our latest master branch?

pillarsdotnet:
@bastelfreak -- Would you please review?

kenyon:
rubocop errors https://github.com/voxpupuli/puppet-archive/actions/runs/3643129527/jobs/6151017200

pillarsdotnet:
@kenyon -- Thanks for fixes.

kenyon (approved):
LGTM, would be nice to have a test for this though.

lib/puppet/provider/archive/ruby.rb
```diff
@@ -201,12 +202,13 @@ def transfer_download(archive_filepath)
...
       if actual_checksum != checksum
         destroy
-        FileUtils.rm_f(temppath) if File.exist?(temppath)
         raise(Puppet::Error, "Download file checksum mismatch (expected: #{checksum} actual: #{actual_checksum})")
       end
     end
 
     move_file_in_place(temppath, archive_filepath)
+  ensure
+    FileUtils.rm_f(temppath) if File.exist?(temppath)
```
smortex, line 211: Technically you can rm_f a file that does not exist (because force is only about ignoring errors). Checking for existence does not hurt though, so I am fine with this.

smortex (approved):
LGTM!  I pushed a commit to make rubocop happy.

PR #482, opened 2023-01-03
Bug fix when passing multiple headers

### LINKED ISSUE #481: curl provider: array of multiple headers does not work

- Module version: **6.1.0**

```puppet
  archive { "example.yml":
    path             => "/tmp/example.yml",
    source           => "https://github.com/examplefile.yml",
    extract_path     => /opt/new/dir,
    headers          => [
      'Accept: application/vnd.github.raw',
      "Authorization: Bearer ${github_token}",
    ],
    provider         => 'curl',
    cleanup          => false,
  }
```

## What are you seeing
Only the first `--header` flag is available at execution
`--header 'Accept: application/vnd.github.raw'`

Repition of headers
`--header 'Accept: application/vnd.github.raw' --header 'Authorization: Bearer xxx_example_token'`

sprankle:
Apprach allowss optional_switch to accept array of values and bulk assign to a flag

Fixes #481

### DIFF
--- a/lib/puppet/provider/archive/ruby.rb
+++ b/lib/puppet/provider/archive/ruby.rb
@@ -259,10 +259,8 @@ Puppet::Type.type(:archive).provide(:ruby) do
   end
 
   def optional_switch(value, option)
-    if value.is_a?(Array)
-      value.map { |item| option.map { |flags| flags % item } }.flatten
-    elsif value
-      option.map { |flags| flags % value }
+    if value
+      Array(value).map { |item| option.map { |flags| flags % item } }.flatten
     else
       []
     end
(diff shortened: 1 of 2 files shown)


kenyon:
I think you should add a test to https://github.com/voxpupuli/puppet-archive/blob/master/spec/unit/puppet/provider/archive/curl_spec.rb which has multiple headers, in a separate PR, so that  we can see the failure, then fix the problem in this PR on top of the other PR, so that we can see the fix.

lib/puppet/provider/archive/ruby.rb
```diff
@@ -259,7 +259,9 @@ def gs_download(path)
   end
 
   def optional_switch(value, option)
-    if value
+    if value.is_a?(Array)
+      value.map { |item| option.map { |flags| flags % item } }.flatten
+    elsif value
       option.map { |flags| flags % value }
```
kenyon, line 265: How about this to eliminate some code duplication:

```suggestion
    if value
      Array(value).map { |item| option.map { |flags| flags % item } }.flatten
```

Tests pass locally with that.

kenyon (approved):
LGTM, should get another review though since some of the code was my contribution.

bastelfreak (approved):

PR #483, opened 2023-01-04
Tests: Curl provider add multi-header request

sprankle:
Add test case when passing multiple headers to the curl provider


kenyon:
Done in #481.

PR #491, opened 2023-04-10
Fix catalog compilation failure when net/ftp is not available

### LINKED ISSUE #488: Missing gem with ruby 3.1

- Puppet: 7.23
- Ruby: 3.1
- Distribution: FreeBSD 
- Module version: All 

Just run the puppet agent, if the module puppet-archive are installed the agent would failed

## What are you seeing

When the puppet agent run (well start) it failed with this message 
```shell
Info: Using environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Error: Could not autoload puppet/provider/archive/ruby: Unable to find archive module in modulepath /usr/local/etc/puppet/modules:/usr/local/etc/puppet/modules
Error: Could not autoload puppet/provider/archive/curl: Could not autoload puppet/provider/archive/ruby: Unable to find archive module in modulepath /usr/local/etc/puppet/modules:/usr/local/etc/puppet/modules
Error: Could not autoload puppet/type/archive: Could not autoload puppet/provider/archive/curl: Could not autoload puppet/provider/archive/ruby: Unable to find archive module in modulepath /usr/local/etc/puppet/modules:/usr/local/etc/puppet/modules
Error: Failed to apply catalog: Could not autoload puppet/type/archive: Could not autoload puppet/provider/archive/curl: Could not autoload puppet/provider/archive/ruby: Unable to find archive module in modulepath /usr/local/etc/puppet/modules:/usr/local/etc/puppet/modules
```
The reason was find by @smortex (big thanks to him) is with ruby 3.1 the net-ftp become a gem (check https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/) the module puppet-archive need this gem. So without this gem installed the agent cannot run

The module puppet-archive should installed the gem so don't need to installed manually on each client (and because puppet agent without puppet-archive don't need net-ftp)

smortex:
When net/ftp is not available, the archive module fail to load and the
PuppetServer catalog compilation fail.  While this gem is generally
available because part of the Ruby standard library, some flavors of
Ruby do noth ship with it by default, e.g. FreeBSD package of Ruby 3.2
or GitHub actions of Ruby 3.2.

Because the archive module has multiple ways to download files, the
absence of net/ftp is not necessarily an issue because when curl(1) is
available, it is preferred over net/ftp.  The catalog compilation error
is therefore not required.

Lazy load net/ftp when the pure-ruby download-over-FTP code is run, so
that the absence of net/ftp cause a resource evaluation error on the
agent side and the rest of the catalog still apply.

Fixes #488

### DIFF
--- a/lib/puppet_x/bodeco/util.rb
+++ b/lib/puppet_x/bodeco/util.rb
@@ -145,9 +145,9 @@ module PuppetX
     end
 
     class FTP
-      require 'net/ftp'
-
       def initialize(url, options)
+        require 'net/ftp'
+
         uri = URI(url)
         username = options[:username]
         password = options[:password]


kenyon (approved):

david22swan (approved):

david22swan:
Any word on this getting merged and released?

ekohl (approved):
> some flavors of
> Ruby do noth ship with it by default, e.g. FreeBSD package of Ruby 3.2
> or GitHub actions of Ruby 3.2.

This isn't "flavors", it's just Ruby 3. In https://github.com/puppetlabs/puppetlabs-apt/pull/1050 I did a similar thing and in https://github.com/puppetlabs/puppetlabs-apt/pull/1050#issuecomment-1448629369 it was suggested to move the require as close as possible to actual use, so I think this is correct.

PR #523, opened 2024-05-28
Resolve 7zip command checks for proper PS fallback

### LINKED ISSUE #364: Windows does not default to extract using powershell

- Puppet:
- Ruby:
- Distribution:
- Module version:

##puppet code
#Enable S3 Support for windows
  class { '::archive':
    aws_cli_install => true,
    seven_zip_provider > 'windows' ##<< Did this to bypass cases that would default to install 7zip in init.pp:11
  }

archive { "zipfile-1.2-windows-x86_64.zip":
              ensure        => present,
              source        => "http://zipfile-1.2-windows-x86_64.zip",
              extract       => 'true',
              path           => "c:/tmp/",
              temp_dir      => "c:/tmp/",
              extract_path  => "c:/Program Files/",
              creates       => "c:/Program Files/zipfile-1.2-windows-x86_64"
          }

## What are you seeing

Error: The source parameter is required when using the Windows provider.
Error: /Stage[main]/Archive/Package[7zip]/ensure: change from absent to present failed: The source parameter is required
 when using the Windows provider.
Debug: Class[Archive]: Resource is being skipped, unscheduling all events
Error: /Stage[main]/Elkagent/Archive[zipfile-1.2-windows-x86_64.zip]/creates: change from archive not extracte
d to extracting in c:/Program Files/ to create c:/Program Files/zipfile-1.2-windows-x86_64 failed: 7z.exe not avail
able

When leaving the provider undefined, it should default to windows powershell if 7zip is not installed.
However, it is still requiring 7zip no matter what options are selected.

klab-systems:
Fixes #364

### DIFF
--- a/lib/puppet_x/bodeco/archive.rb
+++ b/lib/puppet_x/bodeco/archive.rb
@@ -61,12 +61,12 @@ module PuppetX
       private
 
       def win_7zip
-        if ENV['path'].include?('7-Zip') || system('where 7z.exe')
+        if system('where 7z.exe')
           '7z.exe'
-        elsif File.directory?('C:\\Program Files\\7-Zip')
-          'C:\\Program Files\\7-Zip\\7z.exe'
-        elsif File.directory?('C:\\Program Files (x86)\\7-zip')
-          'C:\\Program Files (x86)\\7-Zip\\7z.exe'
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):
not a PS expert, but looks good to me.

